Issue #4, opened 2023-01-05
"write operations are not currently supported, please file a feature request with your use case"


yelhouti:
Hi there,

Great library you have here, sadly it only works for read access. Do you have any plans to support write access ?

Thanks in advance.

oeph:
just for the sake of completeness: this was the discussion at the official java sdk repo https://github.com/aws/aws-sdk-java-v2/issues/1388#issuecomment-1381389190

markjschreiber:
It's great to see people wanting some kind of write support. To support writes, we would need to implement `java.nio.channels.WritableByteChannel` and maybe `java.nio.channels.GatheringByteChannel`.

We could alter the `S3SeekableByteChannel` to implement the `write` and `truncate` calls by delegating to one or more implementations of the above. There are some complications though. S3 doesn't allow random writes so if you attempted to write when the position of the channel was not 0 it would have to fail. This might be OK for some use cases but doesn't generalize well. Pseudo support for random writes could be done if we buffer the writes in memory or in temp files and then only perform the actual write when the channel is closed. For other users of S3 this will be read-after-write consistent. E.g. another S3 client will not see the changes until they are fully written to S3 (which is normal for S3). We could allow consistent reads for a `S3SeekableByteChannel` so that users of that channel see all updates immediately. Harder to implement but probably more what people would expect.

Let me know your thoughts.

oeph:
Would you be up to implement this? I have to say im not an expert, but could try to assist, test, ...

> Pseudo support for random writes could be done if we buffer the writes in memory or in temp files and then only perform the actual write when the channel is closed

Should probably be written to a temp file instead of memory, in order to also support use cases where large files are handled.

> Harder to implement but probably more what people would expect.

True, that could get complicated, but would make sense to do for sure.

markjschreiber:
I can do it but it will certainly be a secondary (perhaps even tertiary) activity.  What kind of time frame are you on? What is your most frequent use case?

oeph:
We are currently trying to get our systems ready and would love to integrate this library sometime around March/April. At least this is the current plan.

For us the most common use case is reading, however we also need to write files, and the file size can vary greatly.
It could range from files of some few megabytes to files of multiple gigabytes.

oeph:
@markjschreiber we got a request to finish this integration sooner than expected. I will try to open a PR in the next days.

markjschreiber:
That would be awesome

[...]

PR #5, opened 2023-02-08
Add NIO write operations

oeph:
**Ticket:** https://github.com/awslabs/aws-java-nio-spi-for-s3/issues/4

*Description of changes:*
* This adds the NIO FileSystemProvider methods as well as a implementation of a writable channel
* With this pr the async s3 client is using now the s3 transfer manager and therefore the crt builder

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ supply the option `StandardOpenOption.READ`. All write operations on the channel
 which will be uploaded to S3 upon closing the channel.
 
 Be aware, that the current implementation only supports channels to be used either for read or write due to potential
-consistency issues we may face in some cases.
+consistency issues we may face in some cases. Attempting to open a channel for both read and write will result in an error.
 
 ### Configuration
 
--- a/README.md
+++ b/README.md
@@ -115,15 +115,31 @@ reduce the number of requests to the S3 service.
 
 Ensure sufficient memory is available to your JVM if you increase the fragment size or fragment number.
 
+## Writing Files
+
+The mode of the channel is controlled with the `StandardOpenOptions`. To open a channel for write access you need to 
+supply the option `StandardOpenOption.READ`. All write operations on the channel will be gathered in a temporary file,
+which will be uploaded to S3 upon closing the channel.
+
+Be aware, that the current implementation only supports channels to be used either for read or write due to potential
+consistency issues we may face in some cases.
+
+### Configuration
+
+The handling of large files could take potentially very long, therefore there are currently no timeouts configured per 
+default. However, you may configure timeouts via the `S3SeekableByteChannel`.
+
+#### Timeouts
+To configure timeouts for writing files or opening files for write access, you may use the `Long timeout` and 
+`TimeUnit timeUnit` parameters of the `S3SeekableByteChannel` constructor.
+```java
+new S3SeekableByteChannel(s3Path, s3Client, channelOpenOptions, timeout, timeUnit);
+```
+
 ## Design Decisions
 
 As an object store, S3 is not completely analogous to a traditional file system. Therefore, several opinionated decisions
 were made to map filesystem concepts to S3 concepts.
-
-### Read Only
-
-The current implementation only supports read operations. It would be possible to add write operations, however special consideration
-would be needed due to the lack of support for random writes in S3 and the read-after-write consistency of S3 objects.
 
 ### A Bucket is a `FileSystem`
 
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ consistency issues we may face in some cases. Attempting to open a channel for b
 
 ### Configuration
 
-The handling of large files could take potentially very long, therefore there are currently no timeouts configured per 
+Because we cannot predict the time it would take to write files, there are currently no timeouts configured per 
 default. However, you may configure timeouts via the `S3SeekableByteChannel`.
 
 #### Timeouts
--- a/src/main/java/software/amazon/nio/spi/s3/S3ClientStore.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3ClientStore.java
@@ -260,11 +260,8 @@ public class S3ClientStore {
         Region region = regionString.equals("") ? Region.US_EAST_1 : Region.of(regionString);
         logger.debug("bucket region is: '{}'", region.id());
 
-        return S3AsyncClient.builder()
+        return S3AsyncClient.crtBuilder()
                 .region(region)
-                .overrideConfiguration(conf -> conf.retryPolicy(builder -> builder
-                        .retryCondition(retryCondition)
-                        .backoffStrategy(backoffStrategy)))
                 .build();
     }
 
--- a/src/main/java/software/amazon/nio/spi/s3/S3SeekableByteChannel.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3SeekableByteChannel.java
@@ -7,61 +7,58 @@ package software.amazon.nio.spi.s3;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import software.amazon.nio.spi.s3.config.S3NioSpiConfiguration;
-import software.amazon.nio.spi.s3.util.TimeOutUtils;
 import software.amazon.awssdk.services.s3.S3AsyncClient;
 import software.amazon.awssdk.services.s3.model.HeadObjectResponse;
+import software.amazon.nio.spi.s3.config.S3NioSpiConfiguration;
+import software.amazon.nio.spi.s3.util.TimeOutUtils;
 
 import java.io.IOException;
 import java.nio.ByteBuffer;
-import java.nio.channels.ClosedChannelException;
-import java.nio.channels.ReadableByteChannel;
-import java.nio.channels.SeekableByteChannel;
-import java.nio.channels.WritableByteChannel;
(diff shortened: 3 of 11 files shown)


oeph:
@markjschreiber could you have a look at this?

src/main/java/software/amazon/nio/spi/s3/S3ClientStore.java
```diff
@@ -260,11 +260,8 @@ private S3AsyncClient asyncClientForRegion(String regionString){
         Region region = regionString.equals("") ? Region.US_EAST_1 : Region.of(regionString);
         logger.debug("bucket region is: '{}'", region.id());
 
-        return S3AsyncClient.builder()
+        return S3AsyncClient.crtBuilder()
```
markjschreiber, line 263: Nice update

src/main/java/software/amazon/nio/spi/s3/S3SeekableByteChannel.java
```diff
@@ -7,61 +7,58 @@
...
 
+    private boolean closed;
     private long size = -1L;
 
-    private final Logger logger = LoggerFactory.getLogger(this.getClass().getName());
+    protected S3SeekableByteChannel(S3Path s3Path, S3AsyncClient s3Client) throws IOException {
+        this(s3Path, s3Client, Collections.emptySet());
+    }
 
-    protected S3SeekableByteChannel(S3Path s3Path, S3AsyncClient s3Client, long startAt) throws IOException{
```
markjschreiber, line 38: Technically this would change the signature of a method that is visible to any subclass someone may have made. Can you use an overloaded method instead and perhaps deprecate this one?

src/main/java/software/amazon/nio/spi/s3/S3SeekableByteChannel.java
```diff
@@ -7,61 +7,58 @@
...
-        S3NioSpiConfiguration config = new S3NioSpiConfiguration();
-        readDelegate = new S3ReadAheadByteChannel(s3Path, config.getMaxFragmentSize(), config.getMaxFragmentNumber(), s3Client, this);
-    }
-
-    /**
-     * Open a new byte channel
-     * @param s3Path the path that the channel will read from
-     * @param startAt the byte offset to start at (0 is the beginning).
-     */
-    protected S3SeekableByteChannel(S3Path s3Path, long startAt) throws IOException {
```
markjschreiber, line 55: Same comment about changing method signatures

src/main/java/software/amazon/nio/spi/s3/S3SeekableByteChannel.java
```diff
@@ -7,61 +7,58 @@
...
-        this(s3Path, S3ClientStore.getInstance().getAsyncClientForBucketName(s3Path.bucketName()), startAt);
-    }
 
-    /**
-     * Equivalent to: {@code new S3SeekableByteChannel(s3Path, 0L)}
-     * @param s3Path the path to open the byte channel for.
-     */
-    protected S3SeekableByteChannel(S3Path s3Path) throws IOException {
```
markjschreiber, line 63: Same comment about changing method signatures.

src/main/java/software/amazon/nio/spi/s3/S3SeekableByteChannel.java
```diff
@@ -7,61 +7,58 @@
...
-        this(s3Path, S3ClientStore.getInstance().getAsyncClientForBucketName(s3Path.bucketName()), 0L);
+        S3NioSpiConfiguration config = new S3NioSpiConfiguration();
+        if (options.contains(StandardOpenOption.WRITE)) {
+            LOGGER.info("using S3WritableByteChannel as write delegate for path '{}'", s3Path.toUri());
+            readDelegate = null;
+            writeDelegate = new S3WritableByteChannel(s3Path, s3Client, options);
+        } else {
```
markjschreiber, line 57: This assumes `StandardOpenOption.READ`? I think we should check that, error if READ and WRITE are selected. Probably error if other options are selected?

src/main/java/software/amazon/nio/spi/s3/S3WritableByteChannel.java
```diff
@@ -0,0 +1,98 @@
...
+
+        try {
+            S3FileSystemProvider fileSystemProvider = (S3FileSystemProvider) path.getFileSystem().provider();
+            boolean exists = fileSystemProvider.exists(client, path);
+
+            if (exists && options.contains(StandardOpenOption.CREATE_NEW)) {
+                throw new FileAlreadyExistsException("File at path:" + path + " already exists");
+            }
+            if (!exists && !options.contains(StandardOpenOption.CREATE_NEW) && !options.contains(StandardOpenOption.CREATE)) {
+                throw new NoSuchFileException("File at path:" + path + " does exist yet");
```
markjschreiber, line 39: does not?

src/main/java/software/amazon/nio/spi/s3/S3WritableByteChannel.java
```diff
@@ -0,0 +1,98 @@
...
+                try (S3TransferManager s3TransferManager = S3TransferManager.builder().s3Client(client).build()) {
+                    s3TransferManager.downloadFile(
+                            DownloadFileRequest.builder()
+                                    .getObjectRequest(GetObjectRequest.builder()
+                                            .bucket(path.bucketName())
+                                            .key(path.getKey())
+                                            .build())
+                                    .destination(tempFile)
+                                    .build()
+                    ).completionFuture().join();
```
markjschreiber, line 53: Can we configure some kind of time out? The danger is that over a slow connection, very large file and/ or other network weirdness this could potentially hang.

src/main/java/software/amazon/nio/spi/s3/S3WritableByteChannel.java
```diff
@@ -0,0 +1,98 @@
...
+        try (S3TransferManager s3TransferManager = S3TransferManager.builder().s3Client(client).build()) {
+            s3TransferManager.uploadFile(
+                    UploadFileRequest.builder()
+                            .putObjectRequest(PutObjectRequest.builder()
+                                    .bucket(path.bucketName())
+                                    .key(path.getKey())
+                                    .build())
+                            .source(tempFile)
+                            .build()
+            ).completionFuture().join();
```
markjschreiber, line 94: I think a time out is also needed here

src/test/java/software/amazon/nio/spi/s3/S3FileSystemProviderTest.java
```diff
@@ -202,7 +281,7 @@ public void checkAccessWhenNoSuchFile() throws IOException, ExecutionException,
         provider.checkAccess(mockClient, foo);
     }
 
-    @Test(expected = UnsupportedOperationException.class)
+    @Test(expected = UnsupportedOperationException.class) // TODO
```
markjschreiber, line 284: Is this todo still required/ valid?

markjschreiber (changes requested):
Looking good. Some improvements suggested.

Importantly, can you also update the README to describe how writes works (using a temp file). How read/ write work together (or not) etc.

oeph:
@markjschreiber thank you for the comments. I will have a look at them today

src/main/java/software/amazon/nio/spi/s3/S3WritableByteChannel.java
oeph: I also thought about this, but what would be a reasonable value for the timeout? I somewhere read that the S3TransferManager will throw an error if the underlying client receives an timeout, but i can't find the reference anymore and don't know how reliable the statement was.

markjschreiber: I agree, it is going to be difficult to set a sensible default without knowing how large an object is and what kind of network connection you have. Perhaps the default should be that there is no time out but it would be good to have a version of these methods that allow a caller to provide a maximum time if they would like to.

markjschreiber:
Don't forget to include updated information in the README about how write works and what users should expect.

oeph:
Yes, I will do this. I just wanted to do the timeout beforehand and update the README as last step.

README.md
```diff
@@ -115,16 +115,32 @@ reduce the number of requests to the S3 service.
...
 Ensure sufficient memory is available to your JVM if you increase the fragment size or fragment number.
 
+## Writing Files
+
+The mode of the channel is controlled with the `StandardOpenOptions`. To open a channel for write access you need to 
+supply the option `StandardOpenOption.READ`. All write operations on the channel will be gathered in a temporary file,
+which will be uploaded to S3 upon closing the channel.
+
+Be aware, that the current implementation only supports channels to be used either for read or write due to potential
+consistency issues we may face in some cases.
```
markjschreiber, line 125:
```suggestion
consistency issues we may face in some cases. Attempting to open a channel for both read and write will result in an error.
```

README.md
```diff
@@ -115,16 +115,32 @@ reduce the number of requests to the S3 service.
...
+
+### Configuration
+
+The handling of large files could take potentially very long, therefore there are currently no timeouts configured per
```
markjschreiber, line 129:
```suggestion
Because we cannot predict the time it would take to write files, there are currently no timeouts configured per 
```

markjschreiber:
@oeph, many thanks for the contribution!

Issue #11, opened 2023-05-20
Open  aws-java-nio-spi-for-s3 Public to S3 compatible providers


stefanofornari:
Hello,
I'd like to use this provider to access S3-like services. To start with, I was thinking of being able to specify the endpoint as part of the s3 url, but another mechanism like passing additional options in the env Map would work as well.
Is anybody already working on something like this? Would you be interested in receiving a contribution?

thanks in advance,
Stefano

markjschreiber:
Nobody is currently working on this. A contribution that could do this would be interesting, especially if it can be done via configuration or composition.

stefanofornari:
Thanks for the reply, I can work on it. What's the best way to discuss the implementation? is this issue ok?

dannycjones:
It might be worth taking a look at how the AWS CLI is providing the option to override the endpoint, as its a similar use case. The CLI is also using `s3://*` URI to provide locations in S3, and uses the `--endpoint-url` argument if the URL needs to be overridden.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/privatelink-interface-endpoints.html#privatelink-aws-cli-examples

markjschreiber:
I think the best way to do this would be to make it easier for a user to provide a customized S3Client using one or more of the following approaches. 

Both the sychronous and asynchronous clients can be built with an endpoint configuration that could include the custom endpoint URL https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/client/builder/AwsClientBuilder.html#withEndpointConfiguration-com.amazonaws.client.builder.AwsClientBuilder.EndpointConfiguration-

1. Currently several public methods on channels have a protected version that allows you to provide a client. These versions are mostly used for testing and allow a Mock client to be used. If they are made public then any client could be used.
2. The `S3ClientStore` class of the library builds a client for each bucket and caches it. You could add methods to allow you to register a different client for a bucket.
3. Similar to 2. you could allow the `S3ClientStore` to be constructed with a semi-configured S3ClientBuilder (and AsyncBuilder) that will be used by default for constructing all clients.

stefanofornari:
Thanks @dannycjones, @markjschreiber. What about one or both of even simpler approaches:

1. add a configuration parameter S3_SPI_ENDPOINT following the pattern used for credentials 
2. allow url like s3://mykey:mysecret@myendpoint.com/mybucket/myresource

[...]

PR #32, opened 2023-06-02
Upstream#11.1

stefanofornari:
[Issue #11](https://github.com/awslabs/aws-java-nio-spi-for-s3/issues/11)

Step 1 in implementing #11 : move the clients cache to S3FileSystemProvider and deprecating singleton usage in S3ClientStore


.gitignore
```diff
@@ -6,6 +6,11 @@ gradle.properties
 
 # Ignore Gradle build output directory
 build
+<<<<<<< HEAD
```
markjschreiber, line 9: Did you mean to change this file?

src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
```diff
@@ -821,4 +825,18 @@ protected Map<String, Object> readAttributes(S3AsyncClient client, Path path, St
     public void setAttribute(Path path, String attribute, Object value, LinkOption... options) throws UnsupportedOperationException {
         throw new UnsupportedOperationException("s3 file attributes cannot be modified by this class");
     }
+
+    /**
+     * Creates if needed and returns the S3ClientStrore caching all created
```
markjschreiber, line 830: Typo in S3ClientStore

src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
```diff
@@ -821,4 +825,18 @@ protected Map<String, Object> readAttributes(S3AsyncClient client, Path path, St
...
+     * clients. This is not public on purpose, as it is an intermediate solution
+     * that will be replaced by instance accessors.
+     *
+     * @return the chache of clients as a S3ClientStore
+     */
+    protected static S3ClientStore getClientStore() {
```
markjschreiber, line 836: If it is intended to be replaced, can it be private?

src/test/java/software/amazon/nio/spi/s3/S3ClientStoreTest.java
```diff
@@ -226,15 +222,16 @@ public void testGenerateAsyncClientWith403Then301ResponsesNoHeader(){
     }
 
     @Test
+    // TODO: move to S3FileSystemProvider
     public void testCaching() {
-        S3Client client = S3Client.create();
+        S3Client client = S3Client.builder().region(Region.US_EAST_1).build();
```
markjschreiber, line 227: Isn't this the same as `S3Client.create()`?

markjschreiber: Why is the region needed?

src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
stefanofornari: It can not be private because it is called instead of S3ClientStore.getInstance() in many places outside the class.

src/test/java/software/amazon/nio/spi/s3/S3ClientStoreTest.java
stefanofornari: In my environment, S3Client.create() throws the following exception:

```
software.amazon.awssdk.core.exception.SdkClientException: Unable to load region from any of the providers in the 
chain software.amazon.awssdk.regions.providers.DefaultAwsRegionProviderChain@721d5b74: [software.amazon.awssdk.regions.providers.SystemSettingsRegionProvider@2b73bd6b: Unable to load region from system settings. Region must be specified either via environment variable (AWS_REGION) or  system property (aws.region)., software.amazon.awssdk.regions.providers.AwsProfileRegionProvider@673e76b3: No region provided in profile: default, software.amazon.awssdk.regions.providers.InstanceProfileRegionProvider@1f66d8e1: Unable to contact EC2 metadata service.]
	at software.amazon.awssdk.core.exception.SdkClientException$BuilderImpl.build(SdkClientException.java:111)
	at software.amazon.awssdk.regions.providers.AwsRegionProviderChain.getRegion(AwsRegionProviderChain.java:70)
	at software.amazon.awssdk.awscore.client.builder.AwsDefaultClientBuilder.regionFromDefaultProvider(AwsDefaultClientBuilder.java:282)
```
I assumed it is because not using AWS I do not have AWS_REGION set. I replaced with the builder since I noticed is used everywhere else.

.gitignore
stefanofornari: Yes, I supposed to add bin and target (the former generated by gradle, the latter by maven)

src/test/java/software/amazon/nio/spi/s3/S3ClientStoreTest.java
markjschreiber: Makes sense to add a region otherwise we are relying on the test environment having a profile or a env var.

.gitignore
markjschreiber: Got it. I am wondering why the `<<<<<` and `>>>>` diff lines are in there though?

Issue #30, opened 2023-06-01
FileSystem always created


stefanofornari:
Is a design choice to create always a new S3FileSystem in both newFileSystem and getFileSystem? It kind of violates the intended behaviour of NIO file systems and should not be a big effort to create a caching mechanism.

I have also noticed that getPath() calls FileSystems.getFileSystem(uri), which then creates a new FileSystem. However S3FileSystem keeps some state like if open or open channels. Could this be a problem?

markjschreiber:
`FileSystem` implementations include a `close()` method which is supposed to be called when done. This method closes all open channels (or should if it doesn't already). It would actually be a good idea to make the implementation implement `AutoClosable` so that it could be used in a "try-with-resources" block. It already has the required method.

One challenge of caching `FileSystem` implementations is that you would then have to carefully consider the behavior if it is used in multiple threads. One thread might close the object while another is trying to use it. What happens if there are two channels to the same object? Sharing them would be really bad. Even without threading if the object is closed, should further requests to the cache return the closed object or create a new one (probably the latter?). You could make the `S3FileSystem` unclosable like the `default` `FileSystem` which would be OK (S3 is never really closed) but we would have to think about how to clean up any channels (probably they would have to be closed individually).

The Java docs for the `FileSystem` interface say that: 
"_File systems are safe for use by multiple concurrent threads. The [close](https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileSystem.html#close--) method may be invoked at any time to close a file system but whether a file system is asynchronously closeable is provider specific and therefore unspecified. In other words, if a thread is accessing an object in a file system, and another thread invokes the close method then it may require to block until the first operation is complete. Closing a file system causes all open channels, watch services, and other [closeable](https://docs.oracle.com/javase/8/docs/api/java/io/Closeable.html) objects associated with the file system to be closed._"

Allowing multiple instances is an easy way to get around this, but perhaps not the best way to do it. I'm open to proposals.

stefanofornari:
I read two topics: closing a FS and multithreading; in both cases, if I got you right, you are concerned that instances of S3FileSystem are not thread-safe and the workaround is creating a new instance when the FS is requested. I think you have a point that we need to take into account the state of a FS when changed potentially concurrently.
However, I do not see (maybe I missed it) how these issues can be prevented by creating always a new instance. In fact I could get an instance calling getFileSystem() only once and access it from multiple threads.
[...]

PR #43, opened 2023-06-06
 Open aws-java-nio-spi-for-s3 Public to S3 compatible providers (Issue #11)

stefanofornari:
https://github.com/awslabs/aws-java-nio-spi-for-s3/issues/11

Step 1&2 in implementing https://github.com/awslabs/aws-java-nio-spi-for-s3/issues/11 : move the clients cache to S3FileSystemProvider and deprecating singleton usage in S3ClientStore

(I had some issues with the branches and at the end I created a new PR, I hope I haven't messed up anything...)


stefanofornari:
With commit [796e0f0](https://github.com/awslabs/aws-java-nio-spi-for-s3/pull/43/commits/796e0f033b7e78530074af2b041b40bab8e60e13) I think I am done with the implementation of #11 . Let me know your thoughts and I will move to  #30

src/test/java/software/amazon/nio/spi/s3/FakeAsyncS3ClientBuilder.java
```diff
@@ -0,0 +1,102 @@
...
+import java.net.URI;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.s3.S3AsyncClient;
+import software.amazon.awssdk.services.s3.S3CrtAsyncClientBuilder;
+import software.amazon.awssdk.services.s3.crt.S3CrtHttpConfiguration;
+import software.amazon.awssdk.services.s3.crt.S3CrtRetryConfiguration;
+
+/**
+ * This class fakes a S3CrtAsyncClientBuilder for testability purposes. It
```
markjschreiber, line 16: I think you should use a Mockito Spy for this purpose (https://www.baeldung.com/mockito-spy)

src/test/java/software/amazon/nio/spi/s3/S3ClientStoreTest.java
```diff
@@ -166,7 +171,7 @@ public void testGenerateAsyncClientWith403Then301Responses(){
 
     @Test
     public void testGenerateClientWith403Then301ResponsesNoHeader(){
-        // when you get a forbidden response from getBucketLocation
+        // when you get a forbidden response from getBucketLocation --- HERE
```
markjschreiber, line 174: Remove `--- HERE`? Not sure what this is for?

src/main/java/software/amazon/nio/spi/s3/S3ClientStore.java
```diff
@@ -197,72 +219,147 @@ protected S3Client generateClient (String bucketName, S3Client locationClient) {
...
+                key = endpoint.substring(0, pos);
+                endpoint = endpoint.substring(pos+1);
+                pos = key.indexOf(':');
+                if (pos >= 0) {
+                    secret = key.substring(pos+1);
+                    key = key.substring(0, pos);
+                }
+            }
+            // TODO: shall we have the protocol in the endpoint already?
+            clientBuilder.endpointOverride(URI.create("https://" + endpoint));
```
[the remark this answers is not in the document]
stefanofornari: this will be fixed later though, when we will support passing preferences in newFileSystem().

src/test/java/software/amazon/nio/spi/s3/FakeAsyncS3ClientBuilder.java
stefanofornari: I have tried to do it, but after a while and for the reason below I parked it. My points are:

1. in general, I prefer using code I can fully control and if possible, not use mocks but the real code, so that I am as closest as possible to the real production code;
2. with the Fake class approach, I am actually using a real BUILDER with all the side effects it may bring
3. As far as I understood spying (although I admit I am not an expert), you can mimic method calls, but you can not inject you own custom logic

src/test/java/software/amazon/nio/spi/s3/S3ClientStoreTest.java
stefanofornari: a left over to be removed

stefanofornari:
I should be done. The commit for JUnit5 is quite big but should be straightforward. Let me know if I have missed anything.

src/main/java/software/amazon/nio/spi/s3/S3ClientStore.java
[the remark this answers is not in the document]
markjschreiber: I'll resolve this comment then but can you add an issue for the TODO so it doesn't get forgotten?

src/test/java/software/amazon/nio/spi/s3/FakeAsyncS3ClientBuilder.java
markjschreiber: A Spy should use a real class (like a Builder). It can record, intercept and modify calls to the real classes methods but perhaps isn't the best way to inject custom logic. What is the custom logic you need?

markjschreiber (approved):
An optional comment about the Spy, it might still be possible to avoid needing a whole new class but not a major problem. Please add the TODO parts about allowing other protocols in endpoint URLS to an existing issue or a new issue so they don't get forgotten.

markjschreiber:
OK to Squash and Merge 👍

src/main/java/software/amazon/nio/spi/s3/S3ClientStore.java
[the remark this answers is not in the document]
stefanofornari: https://github.com/awslabs/aws-java-nio-spi-for-s3/issues/61

src/test/java/software/amazon/nio/spi/s3/FakeAsyncS3ClientBuilder.java
stefanofornari: It's in the functionality of the Fake class: I want to be able to make sure I am passing the right configuration to the real S3Client. As far as I could understand, you can not easily access the config you use in the builder from an instance of a S3Client. I could not even figure out where some parameters where stored, and in some cases I could find them in private fields. A simple use case I wasn't able to figure out how to achieve with a spy:

```
S3ClientStore cs = new S3ClientStore();
cs.generateAsyncClient(...);
assertTrue(URI.create("https://endpoint1.io"), ... client's endpoint... );
```

PR #50, opened 2023-06-11
Adding a S3ClientProvider who generates the clients... (Issue #30)

stefanofornari:
S3ClientStore can implement just the caching/storing logic, while the configuration and creation logic of the clients is delegated to the client provider

*Issue #, if available:*
Issue #30 

*Description of changes:*
S3ClientStore can implement just the caching/storing logic, while the configuration and creation logic of the clients is delegated to the client provider

*Attribution*

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -43,6 +43,7 @@ dependencies {
     // Use JUnit test framework.
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'org.mockito:mockito-core:5.3.1'
+    testImplementation 'com.github.stefanbirkner:system-rules:1.19.0'
 
     implementation platform('software.amazon.awssdk:bom:2.20.79')
 
--- a/build.gradle
+++ b/build.gradle
@@ -43,7 +43,7 @@ dependencies {
     // Use JUnit test framework.
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'org.mockito:mockito-core:5.3.1'
-    testImplementation 'com.github.stefanbirkner:system-rules:1.19.0'
+    testImplementation 'com.github.stefanbirkner:system-lambda:1.2.1'
 
     implementation platform('software.amazon.awssdk:bom:2.20.79')
 
--- a/build.gradle
+++ b/build.gradle
@@ -43,6 +43,8 @@ dependencies {
     // Use JUnit test framework.
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'org.mockito:mockito-core:5.3.1'
+    // todo: to be removed
+    testImplementation 'com.github.stefanbirkner:system-rules:1.19.0'
     testImplementation 'com.github.stefanbirkner:system-lambda:1.2.1'
 
     implementation platform('software.amazon.awssdk:bom:2.20.79')
--- a/src/main/java/software/amazon/nio/spi/s3/S3ClientProvider.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3ClientProvider.java
@@ -82,7 +82,7 @@ public class S3ClientProvider {
         );
     }
 
-    Logger logger = LoggerFactory.getLogger("S3ClientStore");
+    Logger logger = LoggerFactory.getLogger("S3ClientStoreProvider");
 
 
     /**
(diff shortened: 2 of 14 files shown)


src/main/java/software/amazon/nio/spi/s3/S3ClientStore.java
```diff
@@ -34,54 +19,9 @@ public class S3ClientStore {
 
     private static final S3ClientStore instance = new S3ClientStore();
 
-    /**
-     * Default client using the "https://s3.us-east-1.amazonaws.com" endpoint
-     */
-    public static final S3Client DEFAULT_CLIENT = S3Client.builder()
```
markjschreiber, line 40: Removing a public method is a breaking change. I suggest initializing this with the default client of the new S3 provider

src/main/java/software/amazon/nio/spi/s3/S3ClientProvider.java
```diff
@@ -0,0 +1,255 @@
...
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.s3.S3AsyncClient;
+import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.model.HeadBucketResponse;
+import software.amazon.awssdk.services.s3.model.S3Exception;
+
+/**
+ *
+ */
+public class S3ClientProvider {
```
markjschreiber, line 38: Would it be better to have an interface with different implementations for different S3 providers? I'm not sure how different we would expect them to be so maybe this is not needed?

build.gradle
```diff
@@ -43,6 +43,7 @@ dependencies {
     // Use JUnit test framework.
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'org.mockito:mockito-core:5.3.1'
+    testImplementation 'com.github.stefanbirkner:system-rules:1.19.0'
```
markjschreiber, line 46: I will need to check if we can include a dependency on a library that isn't MIT or Apache-2 licensed.

markjschreiber: If it is allowed you will also need to update the `THIRD-PARTY` file

markjschreiber: Have you considered using https://github.com/stefanbirkner/system-lambda? It is MIT licensed. Also takes advantage of Java 8+ features.

src/main/java/software/amazon/nio/spi/s3/S3ClientProvider.java
stefanofornari: I would say not needed for now. I'd like to keep the integration surface at the Was client (i.e. the suppliers shall be compliant with the S3 API. Using a different client would mean fundamentally write a new FileSystem.

build.gradle
stefanofornari: Good catch! I used it as well in another project. I am going to switch to system-lambda.

src/main/java/software/amazon/nio/spi/s3/S3ClientStore.java
```diff
@@ -92,6 +32,8 @@ private S3ClientStore(){}
      */
     public static S3ClientStore getInstance() { return instance; }
 
+    protected S3ClientProvider provider = new S3ClientProvider();
```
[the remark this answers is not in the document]
stefanofornari: I am not sure about this, but I am open. The reason is that I would not provide a public interface to the provider so not to have future legacy. Eventually S3ClientStore will be deprecated/not used any more. As a general rule, providing setters/getters that simply manipulate an instance is useless code (I know here I am going against the holy principle of encapsulation :) ). Having said that, if you still prefer to have it, I will add the constructor.

build.gradle
stefanofornari: I remember why I haven't used it :) it requires JUnit5. For Junit4 we need to use System Rules. Are you ok muve to JUnit5?

markjschreiber: JUnit5 should be OK. Maybe even preferable.

stefanofornari:
I am committing this as a first draft. Main changes:

1. S3FileSystemProvider now creates and caches S3FileSystems based on #30 (the key for the cache is described in https://github.com/awslabs/aws-java-nio-spi-for-s3/issues/30#issuecomment-1595667542)
2. Implement the newFileSystem/getFileSystem semantic
3. closing a FS now makes sure the provider removes it prom the cache
4. using provider.newFileSystem in test classes where new S3FileSystem() was used (this required to add mocks)

stefanofornari:
@markjschreiber, I think we can close this one

Issue #66, opened 2023-06-20
directory stream provides all keys (but broken) when listing a directory


stefanofornari:
**To reproduce:**

Prepare a bucket with the following objects:

```
0000002160133698.pdf
subdir1/index.jpeg
subdir1/subdir11/targets.jpg
info.txt
```
Run the below code:

```
Path path = s3.getPath("/");
Files.list(path).forEach((p) -> System.out.println(p));
```

**Expected result:**
0000002160133698.pdf
subdir1
info.txt

**Actual result:**
0000002160133698.pdf
subdir1
subdir1
info.txt

stefanofornari:
After deeper analysis, I realized the issue was different and is due to the fact that in 

https://github.com/awslabs/aws-java-nio-spi-for-s3/blob/a8e88d2d30319c0506c1ddd10e3b609b3f0b5bc4/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java#L270C1

we first retrieve all keys and then apply a filter on the prefix without actually deduping the result. Please see PR #76 for the fix.

However, the big issue here is retrieving all keys and then determine the file list on the client, it would break quickly with even small file systems. Accordingly to the documentation (https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html), we should make use of the separator to let the server make the split between files and directories. This complicates a bit the processing of the response because files and dirs are going to be in two different sections of the result, but I guess it is a forced choice.

@markjschreiber, can you confirm? would the core team take care of it?

stefanofornari:
We should also implement the MaxKey/Truncate logic

stefanofornari:
I am reopening this one as it is still not working.
Use case:

```
for (Path root : s3.getRootDirectories()) {
    Stream<Path> stream = Files.walk(root);
    for (Path path: stream.filter(Files::isRegularFile).toList()) {
        System.out.println(path);
    }
}
```
The fix should be easy though, just remove line:

https://github.com/awslabs/aws-java-nio-spi-for-s3/blob/ea1044d51bb4f61af05b6651539a719b7a03c529/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java#L290

(the returned paths shall be absolute to allow walking the tree).

stefanofornari:
Pagination seems to be working, great job. The way you solved it and with such small code is remarkable!
Do we need maybe some more tests?

PR #74, opened 2023-06-24
Quick fix for issue #66

stefanofornari:
Issue #66 

*Description of changes:*

Added .distinct() to avoid duplicates in the returned list. This is a quickfix just to move on with the development but there is a corner case I believe it is not covered when there are two keys one representing a file and one a directory. However please see the note below (that I will report to the issue as well).

If I well understood, the current implementation retrieves all keys (!) in the bucket and then filters the one matching the provided prefix. This is highly undesirable as it would break for buckets of a noticeable size. See Isse #66 for more.

### DIFF
--- a/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
@@ -274,6 +274,7 @@ public class S3FileSystemProvider extends FileSystemProvider {
                     .contents()
                     .stream()
                     .map(s3Object -> truncateByPrefix(fs, prefix, s3Object))
+                    .distinct()
                     .filter(path -> {
                         try {
                             return filter.accept(path);


stefanofornari:
@markjschreiber can we merge this one?

markjschreiber:
Looks good. Can you also update the unit test to ensure that future changes don't regress?

> If I well understood, the current implementation retrieves all keys (!) in the bucket and then filters the one matching the provided prefix. 
No, the call the the `ListObjectsV2` supplies a bucket name and a prefix. What comes back from S3 is paths from the bucket that match the prefix (effectively server side filtered). However, this could still be a very large list but there isn't a lot we can do about that.

>but there is a corner case I believe it is not covered when there are two keys one representing a file and one a directory
Hopefully that is prevented by our `PosixLikePath` rules which mean the `foo` is and object and `foo/` is a directory. This is a bit different from true POSIX where if `foo/` is a directory then so it `foo`.

stefanofornari:
> Looks good. Can you also update the unit test to ensure that future changes don't regress?

 Sure, can you just help me out with building a mock to support this case? I am not very familiar with mockito :-1: 

> > If I well understood, the current implementation retrieves all keys (!) in the bucket and then filters the one matching the provided prefix.
> 
> No, the call the the `ListObjectsV2` supplies a bucket name and a prefix. What comes back from S3 is paths from the bucket that match the prefix (effectively server side filtered). However, this could still be a very large list but there isn't a lot we can do about that.

I guess that when prefix is just / (or "") it retrieves everything. Anyway as you said a subdir as well can contain a huge amount of keys. Please have a look at #66, I pointed to the documentation of the ListObjectsV2 call. The way to address it is using also `.separator()` . 
Would be willing to take care of it?

markjschreiber:
Agree. I was just talking with a colleague and using `.delimiter("/")` in the API call should reduce problems with large numbers of keys. It might also remove the need for using `distinct` and the possible edge case where objects and directories have the same name as well but I'm not sure. 

Another design consideration is that the `listObjectsV2` will return up to 1000 objects so we might need to add some logic to make multiple calls if there are more to return.

The general idea with Mockito is that there is a mocked S3Client. In this mocked client you define the implementation of certain methods that you want to test so that when they are called the mock returns some test value rather than making a real call to S3. So you can setup a test to ensure you correctly handle duplicates by making a mock client that returns a list with duplicates when `listObjectsV2` is called. Or a mock where the response from the client indicates that more values remain allowing you to test the logic required to make multiple calls.

I may get to this today but probably not until tomorrow.

src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
```diff
@@ -274,6 +274,7 @@ protected DirectoryStream<Path> newDirectoryStream(S3AsyncClient s3Client, Path
                     .contents()
                     .stream()
                     .map(s3Object -> truncateByPrefix(fs, prefix, s3Object))
```
markjschreiber, line 276: If we add `.delimiter("/")` to the API call then we probably don't need this map or the `truncateByPrefix` private method?

stefanofornari: Probably we need the map because we want to return S3Path objects, but yes, I think truncateByPrefix() won't be needed.

stefanofornari:
> Another design consideration is that the `listObjectsV2` will return up to 1000 objects so we might need to add some logic to make multiple calls if there are more to return.

Yes, we need to use [continuation-token](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html#API_ListObjectsV2_RequestSyntax)


> The general idea with Mockito is that there is a mocked S3Client. In this mocked client you define the implementation of certain methods that you want to test so that when they are called the mock returns some test value rather than making a real call to S3. So you can setup a test to ensure you correctly handle duplicates by making a mock client that returns a list with duplicates when `listObjectsV2` is called. Or a mock where the response from the client indicates that more values remain allowing you to test the logic required to make multiple calls.
> 
> I may get to this today but probably not until tomorrow.

Thanks! an example would be very helpful.

markjschreiber:
@stefanofornari, can you take a look at  this branch https://github.com/awslabs/aws-java-nio-spi-for-s3/tree/markjschreiber/issue66

I think it resolves the issue with Directory streams. It now uses the `delimiter("/")` method to prevent listing all keys and it also uses a paginator for handling more than 1000 entries. I need to test this more but would be good if you can try it out too.

stefanofornari:
Hi Mark, see https://github.com/awslabs/aws-java-nio-spi-for-s3/issues/75#issuecomment-1616076980, the first test on MinIO are not successful.

On the code:

https://github.com/awslabs/aws-java-nio-spi-for-s3/blob/b54d783f15e5c65a4d3fceb30c9ba032ad6464a2/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java#L293

What about just use `substring(FinalPathString.length()) `so to avoid the overheard of regex?

About this snippet:

https://github.com/awslabs/aws-java-nio-spi-for-s3/blob/b54d783f15e5c65a4d3fceb30c9ba032ad6464a2/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java#L285-L301

I could not try it, but I have the feeling it still get all keys before creating the stream, which takes us at square 0,

markjschreiber:
The `delimiter` method of the builder means that the call to S3 will only return 1 level of objects and directories (common prefixes) rather than the full bucket content.

stefanofornari:
> The `delimiter` method of the builder means that the call to S3 will only return 1 level of objects and directories (common prefixes) rather than the full bucket content.

True, but still one level can have thousands elements. We probably need to build an iterator that asks one page at a time.

markjschreiber:
The call in the new code is wrapped in an autopaginator

stefanofornari:
> The call in the new code is wrapped in an autopaginator

Yes of course, but you are collecting all responses first, then you create the iterator on s3Paths.

markjschreiber:
True. I probably need to use a join iterator that lazily combines the commonPrefixes and objects returned in each page triggering new pages when it runs out of items. Even better would be to buffer them in the background if we assume that iteration over the whole directory is likely.

stefanofornari:
Yep exactly. I would keep the lazy approach since network is expensive and it may not be needed. It would be easy to iterate through all results in a separate thread by the caller if needed.

markjschreiber:
Complete fix by merge 92

Issue #61, opened 2023-06-17
File System settings


stefanofornari:
AS A user
I WANT to pass file system configuration at file system creation (i.e. FileSystems.newFileSystem())
SO THAT I can set the run time behaviour of the file system

NOTES:

1. we can use aws.region, aws.accessKey and aws.accessKeySecret for region and credentials
2. we could introduce aws.protocol = http/https to specify which protocol to use for endpoint different than QWS S3
3. we could introduce aws.endpoint to specify the endpoint to use if not otherwise given
4. properties passed to newFileSystem() overwrites environment variables and system properties
5. properties passed to newFileSystem() are overwritten by the values provided in URIs

@markjschreiber can you please provide feedback on the notes above?

stefanofornari:
Thanks, this helps. There is only one bullet I am not sure I have got it right. bullet 2. URIs to access a bucket even in the case of an endpoint would be something like:

`s3://endpoint.somewhere.com/bucket`

The scheme/protocol is going to to be s3 anyway.  Are you suggesting we should use s3://https://endpoint.somewhere.com/bucket ? This would probably not be parsed as a URI.

markjschreiber:
Can you give an example of an endpoint that you are wanting to process where you need to determine the protocol?

dannycjones:
> Thanks, this helps. There is only one bullet I am not sure I have got it right. bullet 2. URIs to access a bucket even in the case of an endpoint would be something like:
> 
> `s3://endpoint.somewhere.com/bucket`
> 
> The scheme/protocol is going to to be s3 anyway. Are you suggesting we should use s3://https://endpoint.somewhere.com/bucket ? This would probably not be parsed as a URI.

Does it make more sense to keep endpoint as a separate parameter?

For example, configuring an environment variable like this:

```
S3_SPI_ENDPOINT=https://endpoint.somewhere.com/
```

Then, just go ahead and use the URI `s3://bucket/`.

If you needed to use HTTP rather than HTTPS, then that would be configurable in a consistent way with other S3 tooling. Of course, using HTTPS is recommended. :)

---

This is similar to AWS CLI where you can set the endpoint separately using a CLI option:

```
aws s3 --endpoint-url https://endpoint.somewhere.com ls s3://bucket/
```

Or [Apache Hadoop's S3A connector](https://hadoop.apache.org/docs/stable/hadoop-aws/tools/hadoop-aws/) using a system property:

```
[...]

Issue #40, opened 2023-06-06
Better method to derive bucket name from URI


markjschreiber:
The current method to determine a bucket name from a URI involves parsing the authority from the URI. https://github.com/awslabs/aws-java-nio-spi-for-s3/blob/main/src/main/java/software/amazon/nio/spi/s3/S3FileSystem.java#L58

This probably won't work for URIs that use S3 Access Points, S3 multiregion access points and possibly not s3 buckets on AWS Outposts or Snowball devices.

```[tasklist]
### Tasks
- [ ] Check Access Points
- [ ] Check MR access points
- [ ] Check Outposts
- [ ] Check Snowball devices
- [ ] Build unit test examples of all of the above
```

stefanofornari:
in addtion, PR #43, parses for uri in the form of s3://[[key:secret]@endpoint]/bucket/resource

stefanofornari:
This is going to be address with #30; I won't be able to check it with with mentioned tasks, hopefully @markjschreiber you can do it.

markjschreiber:
We don't need to test against physical devices. The main thing is to check the specification of what the URI for any of the above would look like and if it is different from the normal S3 URI then ensure the logic to get the bucket name can account for this. The patterns for the URIs should all be available from the docs for these products.

dannycjones:
> in addtion, PR #43 (former #32), parses for uri in the form of s3://[[key:secret@]endpoint]/bucket/resource

I'd recommend to have the S3 URI as only bucket and key for now.

Keep endpoint to configuration and credentials to configuration as well, if it cannot already be met by the [default credential provider chain](https://docs.aws.amazon.com/sdkref/latest/guide/standardized-credentials.html#credentialProviderChain) which includes environment variable configuration.

stefanofornari:
We have a problem with this approach... The FileSystem API provides a newFileSystem(URI, Map) but only getFileSystem(URI). When we cache the file system created in newFileSystem(), we have to use endpoint/bucket as key for the cache because we may have two file systems with same bucket name but on different endpoints. But if we do not have a way to give the endpoint to getFileSystem() we won't be able to retrieve the cached instance.

We could work around it ignoring S3_SPI_ENDPOINT passed in env, but then the following won't work:

```
Properties p = System.getProperties();
p.setProperty(S3_SPI_ENDPOINT, "http://endpoint1.com");

[...]

PR #77, opened 2023-06-25
Issues #11, #30, #61 (and #5)

stefanofornari:
*Issue #, if available:*
#11, #30, #61, #5

*Description of changes:*
This is the combined code for issues #11, #30 and #61

#11 - Be able to use compatible providers by supporting 3rd party endpoints
#30 - new S3FileSystemProvider/S3FileSystem to better support #11 and implementing the FileSystems's semantic as per  JDK documentation (i.e. new/get file system); please see the comments in the issue as there are many considerations and feedback
#61 - Be able to configure a file system passing environment properties in newFileSystem() - still work in progress, see comments in the issue
#5 - A side effect of #11 and #61


stefanofornari:
I am not sure why gradle is complaining...

stefanofornari:
@markjschreiber , I consider this PR complete now. I am pretty happy of the result and I can now connect to a S3-like service like MinIO. There is still quite a bit work to do to support all NIO file system functionality, but we are at a great place now.
It would be quite important to integrate this PR so that I can test the core functionalities with real S3-like services.

marco-luzzara:
> I am not sure why gradle is complaining...

I think it is because you are using `record`s, which are a Java 14 preview feature, and stable from Java 16. In the `gradle.build`, you have this:

```
compileJava {
    // see https://docs.gradle.org/6.6/userguide/building_java_projects.html#sec:java_cross_compilation
    options.release = 8
}
```

Upgrading it to `options.release = 16` would be enough to compile it. However, it would probably go against the [Contribution guidelines](https://github.com/awslabs/aws-java-nio-spi-for-s3#contributing) because 

> Code must compile to JDK 1.8 compatible bytecode

marco-luzzara:
@stefanofornari I am looking at your work, thank you very much for all the effort. 

I have cloned you `main-upstream` branch and I have seen that there is no property `s3.spi.endpoint`, as you discussed [here](https://github.com/awslabs/aws-java-nio-spi-for-s3/issues/61#issuecomment-1597607734). Only the endpoint protocol is configurable (`s3.spi.endpoint-protocol`), why?

Moreover, I would like to suggest an alternative approach for the configurations setting. I was inspired by a [FileSystemProvider for the SFTP protocol](https://github.com/robtimus/sftp-fs). In this library, additional configuration are specified through the [`SFTPEnvironment` class](https://github.com/robtimus/sftp-fs/blob/master/src/main/java/com/github/robtimus/filesystems/sftp/SFTPEnvironment.java). The advantage is that this class implements the `Map<String, Object>` interface and the user does not have to specify the property names, but there are utility methods that make the initialization much more readable. 

Back to `S3NioSpiConfiguration`, instead of doing this:

```
FileSystems.newFileSystem(myuri, Map.of(
    AWS_ACCESS_KEY_PROPERTY, "access-key",
    AWS_SECRET_ACCESS_KEY_PROPERTY, "secret-key",
    S3_SPI_ENDPOINT_PROTOCOL_PROPERTY, "http"
));
```

I would like to do:

```
FileSystems.newFileSystem(myuri, new S3NioSpiConfigurationBuilder()
    .withAccessKey("access-key")
    .withSecretKey("secret-key")
    .withEndpointProtocol("http")
);
```

Where the `S3NioSpiConfigurationBuilder` class implements the map interface, of course.

What do you think @markjschreiber @stefanofornari ?

markjschreiber:
> > I am not sure why gradle is complaining...
> 
> I think it is because you are using `record`s, which are a Java 14 preview feature, and stable from Java 16. In the `gradle.build`, you have this:
> 
> ```
> compileJava {
>     // see https://docs.gradle.org/6.6/userguide/building_java_projects.html#sec:java_cross_compilation
>     options.release = 8
> }
> ```
> 
> Upgrading it to `options.release = 16` would be enough to compile it. However, it would probably go against the [Contribution guidelines](https://github.com/awslabs/aws-java-nio-spi-for-s3#contributing) because
> 
> > Code must compile to JDK 1.8 compatible bytecode

Yes, the JDK used in the github action doesn't know about `record` and if we can preserve compatibility with 1.8 or if we have to JDK 11 that would be ideal.

markjschreiber:
I'm wondering if we can break this up into distinct pull requests? 

* There are some parts that fix fundamental issues (Filesystem isn't a singleton but should be). 
* Then there is the need to be able to configure the endpoint url/ protocol.
* There is the possibility of being able to support third party S3 providers (this requires the first two) but is also a new capability on top of them. I *think* that the first two parts would be enough to support openstack but perhaps not minio??
* There may be a need to move to a more recent version of the JDK. This is currently breaking the build of this PR. If there is a need this should be isolated to the minimal PR that needs it.

This will be easier to progress as distinct parts,

stefanofornari:
> > > I am not sure why gradle is complaining...
> > 
> > 
> > I think it is because you are using `record`s, which are a Java 14 preview feature, and stable from Java 16. In the `gradle.build`, you have this:
> > ```
> > compileJava {
> >     // see https://docs.gradle.org/6.6/userguide/building_java_projects.html#sec:java_cross_compilation
> >     options.release = 8
> > }
> > ```
> > 
> >     
> >   
> > Upgrading it to `options.release = 16` would be enough to compile it. However, it would probably go against the [Contribution guidelines](https://github.com/awslabs/aws-java-nio-spi-for-s3#contributing) because
> > > Code must compile to JDK 1.8 compatible bytecode
> 
> Yes, the JDK used in the github action doesn't know about `record` and if we can preserve compatibility with 1.8 or if we have to JDK 11 that would be ideal.

Thanks @marco-luzzara for the feedback and suggestion.
@markjschreiber , are you referring to github actions? I see [here](https://github.com/actions/setup-java)  that modern JDKs are supported as well. Java 8 is a dinosaur of 2014, let's move on please :) )

stefanofornari:
> I have cloned you `main-upstream` branch and I have seen that there is no property `s3.spi.endpoint`, as you discussed [here](https://github.com/awslabs/aws-java-nio-spi-for-s3/issues/61#issuecomment-1597607734). Only the endpoint protocol is configurable (`s3.spi.endpoint-protocol`), why?

Yeah, at the end I thought it could be misleading because it won't be read from the environment, but adding it, it's not a big deal. I'll add a user story for it.



> What do you think @markjschreiber @stefanofornari ?

I like it. Maybe not implementing Map... what about new S3NioSpiConfiguration().withXXX().toMap() ?

stefanofornari:
> I'm wondering if we can break this up into distinct pull requests?

Tricky but I can do it if it speeds up things. A few clarifications below:

> * There are some parts that fix fundamental issues (Filesystem isn't a singleton but should be).

I am not sure I've got this one. I have done a lot of work to make S3FileSystem not singleton. The provider creates/reuses instances given the configuration and the url in newFileSystem() and getFileSystem().

> * Then there is the need to be able to configure the endpoint url/ protocol.
> 
> * There is the possibility of being able to support third party S3 providers (this requires the first two) but is also a new capability on top of them. I _think_ that the first two parts would be enough to support openstack but perhaps not minio??

These to me are only one; we can't support other providers if we do not have the endpoint in the URL (actually, we could if we add the endpoint as configuration, but I am wondering if it is worth - but let me know, if you prefer that, I can do it).
I haven't got though the distinction you made between openstack and minio, can you please elaborate it a bit?


> * There may be a need to move to a more recent version of the JDK. This is currently breaking the build of this PR. If there is a need this should be isolated to the minimal PR that needs it.

this should be easy, ok

marco-luzzara:
> > I have cloned you `main-upstream` branch and I have seen that there is no property `s3.spi.endpoint`, as you discussed [here](https://github.com/awslabs/aws-java-nio-spi-for-s3/issues/61#issuecomment-1597607734). Only the endpoint protocol is configurable (`s3.spi.endpoint-protocol`), why?
> 
> Yeah, at the end I thought it could be misleading because it won't be read from the environment, but adding it, it's not a big deal. I'll add a user story for it.

Thank you. *it won't be read from the environment*, what do you mean? I prefer the `s3.spi.endpoint` because I think it is more straightforward for the user. Supporting both the `s3://my.endpoint:port/bucket/path` and `s3://bucket/path` seems unnecessary to me compared to having a standardized format (which is `s3://bucket/path`) and the optional  `s3.spi.endpoint`. By the way, if we can implement it by the next release, it would be great (I can contribute as well).
> 
> > What do you think @markjschreiber @stefanofornari ?
> 
> I like it. Maybe not implementing Map... what about new S3NioSpiConfiguration().withXXX().toMap() ?

As long as I have the means to configure it easily, it is perfectly fine for me. Thanks again

stefanofornari:
> > > I have cloned you `main-upstream` branch and I have seen that there is no property `s3.spi.endpoint`, as you discussed [here](https://github.com/awslabs/aws-java-nio-spi-for-s3/issues/61#issuecomment-1597607734). Only the endpoint protocol is configurable (`s3.spi.endpoint-protocol`), why?
> > 
> > 
> > Yeah, at the end I thought it could be misleading because it won't be read from the environment, but adding it, it's not a big deal. I'll add a user story for it.
> 
> Thank you. _it won't be read from the environment_, what do you mean? I prefer the `s3.spi.endpoint` because I think it is more straightforward for the user. Supporting both the `s3://my.endpoint:port/bucket/path` and `s3://bucket/path` seems unnecessary to me compared to having a standardized format (which is `s3://bucket/path`) and the optional `s3.spi.endpoint`. By the way, if we can implement it by the next release, it would be great (I can contribute as well).

You can refer to this comment: https://github.com/awslabs/aws-java-nio-spi-for-s3/issues/40#issuecomment-1605740100

I would support

`FileSystems.newFileSystem("s3://mybucket", Map.of("s3.spi.endpoint", "mys3")`)

because it is an instance value, but not
```

System.getProperties().setProperty("s3.spi.endpoint", "mys3");
FileSystems.newFileSystem("s3://mybucket");
```

because of the mentioned comment.

markjschreiber:
FYI, I had to roll back your previous commit as the parsing of the `s3://server:port/bucket/key` pattern broke the standard `s3://bucket/path/to/object` causing the SPI to think that the bucket name was `path`. 

I suspect distinct provider classes  will be needed for s3 compatible services using alternative patterns. 

By the way, I was looking at minio docs and didn't find any mention of s3 URIs being used at all. I did find the use of aliases (which won't help). Is the `s3://server:port/bucket/key` pattern standard for minio??

stefanofornari:
> FYI, I had to roll back your previous commit as the parsing of the `s3://server:port/bucket/key` pattern broke the standard `s3://bucket/path/to/object` causing the SPI to think that the bucket name was `path`.

That's strage as we shoud have a test that fails in such case. Did any capture the issue? If not, can you please provide a test that breaks?


> I suspect distinct provider classes will be needed for s3 compatible services using alternative patterns.

Let's see what the problem is, all test were passing, maybe it was something with the merge?


> By the way, I was looking at minio docs and didn't find any mention of s3 URIs being used at all. I did find the use of aliases (which won't help). Is the `s3://server:port/bucket/key` pattern standard for minio??

I never said it is a standard. It is just natural to use such a url schema because all developers are used to have scheme://endpoint/resource for API calls. It is a convenient way to give the provider the endpoint to use and the users an straightforward  to refer a bucket whenever service owns it.

markjschreiber:
> > FYI, I had to roll back your previous commit as the parsing of the `s3://server:port/bucket/key` pattern broke the standard `s3://bucket/path/to/object` causing the SPI to think that the bucket name was `path`.
> 
> That's strage as we shoud have a test that fails in such case. Did any capture the issue? If not, can you please provide a test that breaks?

The `bucketName` method in `S3PathTest` has been updated to give better coverage and ensures the correct bucket name is retrieved from a URI string

> 
> > I suspect distinct provider classes will be needed for s3 compatible services using alternative patterns.
> 
> Let's see what the problem is, all test were passing, maybe it was something with the merge?
> 

The problem was with the split on `/` and the assumption that if there was more than 1 part then the first part would be the endpoint and the second would be the bucket. The original `S3PathTest` wasn't adequate to show this break. The new version is better.

> > By the way, I was looking at minio docs and didn't find any mention of s3 URIs being used at all. I did find the use of aliases (which won't help). Is the `s3://server:port/bucket/key` pattern standard for minio??
> 
> I never said it is a standard. It is just natural to use such a url schema because all developers are used to have scheme://endpoint/resource for API calls. It is a convenient way to give the provider the endpoint to use and the users an straightforward to refer a bucket whenever service owns it.

Unfortunately the standard S3 URI doesn't contain any information about the endpoint so it would not be easy to differentiate a URI with an endpoint from one without. You can't even make use of the presence of `:` as some specialized S3 URIs include this.

You could fork this repository and use a difference scheme like `minio://` and then change the `SCHEME` constant in `S3FileSystemProvider` to `minio`. Most of the rest of the package would then "just work".

However, I do see that minio says that it is compatible with standard S3 tools so presumably it is possible to do the equivalent of `aws s3 cp my_local_file s3://my_bucket/my/destination/`. What would the URI pattern be in that case?

stefanofornari:
> > > FYI, I had to roll back your previous commit as the parsing of the `s3://server:port/bucket/key` pattern broke the standard `s3://bucket/path/to/object` causing the SPI to think that the bucket name was `path`.
> > 
> > 
> > That's strage as we shoud have a test that fails in such case. Did any capture the issue? If not, can you please provide a test that breaks?
> 
> The `bucketName` method in `S3PathTest` has been updated to give better coverage and ensures the correct bucket name is retrieved from a URI string
> 
> > > I suspect distinct provider classes will be needed for s3 compatible services using alternative patterns.
> > 
> > 
> > Let's see what the problem is, all test were passing, maybe it was something with the merge?
> 
> The problem was with the split on `/` and the assumption that if there was more than 1 part then the first part would be the endpoint and the second would be the bucket. The original `S3PathTest` wasn't adequate to show this break. The new version is better.
> 
> > > By the way, I was looking at minio docs and didn't find any mention of s3 URIs being used at all. I did find the use of aliases (which won't help). Is the `s3://server:port/bucket/key` pattern standard for minio??
> > 
> > 
> > I never said it is a standard. It is just natural to use such a url schema because all developers are used to have scheme://endpoint/resource for API calls. It is a convenient way to give the provider the endpoint to use and the users an straightforward to refer a bucket whenever service owns it.
> 
> Unfortunately the standard S3 URI doesn't contain any information about the endpoint so it would not be easy to differentiate a URI with an endpoint from one without. You can't even make use of the presence of `:` as some specialized S3 URIs include this.
> 
> You could fork this repository and use a difference scheme like `minio://` and then change the `SCHEME` constant in `S3FileSystemProvider` to `minio`. Most of the rest of the package would then "just work".
> 
> However, I do see that minio says that it is compatible with standard S3 tools so presumably it is possible to do the equivalent of `aws s3 cp my_local_file s3://my_bucket/my/destination/`. What would the URI pattern be in that case?

yeah, I checked that and accordingly to https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html the bucket name can not contain ':', therefore parsing the url is doable and it is implemented already in S3URI (we may want to change the name of the class if we are disturbed by the fact it does not represent a real S3 URI.


> However, I do see that minio says that it is compatible with standard S3 tools so presumably it is possible to do the equivalent of `aws s3 cp my_local_file s3://my_bucket/my/destination/`. What would the URI pattern be in that case?

I would say, as mentioned [here](https://github.com/awslabs/aws-java-nio-spi-for-s3/issues/61#issuecomment-1597607734), something like

`aws s3 --endpoint-url https://myminio.somewhere.com cp my_local_file s3://bucket/my_bucket/my/destination/`

Which actually shows, outside the AWS world, how much cumbersome it is and how different pattern is than what developers and users are used to do. We are accustomed to:

```
cp dir1/fie1 /volume2/dir2/file2 (not cp -volume volume2 dir1/file1 
scp dir1/fie1 /volume2/dir2/file2 (not host:/volume2/dir2/file2)
curl https://host1/volume1/file1 (not curl -host host1 https://volume1/file1)
curl sftp://host1/volume1/file1 (not curl -host host1 sftp://volume1/file1)
...
```

Also note that Paths.getPath() would not work if you do not have a way to provide the endpoint in the URI (reminder: using system properties does not work either in NIO world).

I really do not see why this resistance to the provided implementation... maybe I am missing something. If so, can you please elaborate on the practical downsides of this solution?

stefanofornari:
> I'm wondering if we can break this up into distinct pull requests?

There was also quite a bit of work to move to JUnit 5. I think we should do this first, otherwise merging on my side is painful. If I provide the PR, can you integrate it quickly?

markjschreiber:
Adding back JUnit5 would be a good distinct PR. It was something I wanted to avoid with the rollback but it made the merge very messy.

markjschreiber:
> I really do not see where this resistance comes from... maybe I am missing something. If so, can you please elaborate on the practical downsides of this solution?

The problem I see is that the S3 URIs don't contain any information that define an endpoint. The endpoint used is chosen by a client (potentially using configuration information). Even for a standard S3 URI on Amazon S3 there are potentially multiple endpoints that can be used (standard, FIPS, dual stack). This project doesn't own the definition of S3 URIs so if we were to invent a convention where we put endpoint information into an S3 URI we are introducing a convention that doesn't match what the Amazon S3 team has done. If they were to add new patterns in the future it may break any unofficial convention we adopt. 

I do think we might be able to solve this problem in a different way. These are the possible scenarios that we would ideally be able to support.

1. `java MyApp s3://mybucket/path` where `mybucket` is in AWS. This is currently supported.
2. `java MyApp s3://mybucket/path` where `mybucket` is not in AWS, or where it is in AWS but we want to use an alternative endpoint like FIPS endpoint. This can be solved if we provide a way to change the default endpoint for S3 clients.
3. `java MyApp s3://mybucket1/path s3://mybucket2/path`. If `mybucket1` and `mybucket2` need different endpoints we need a mechanism to configure a clients endpoint for each bucket
4. `java MyApp s3://mybucket/path s3://mybucket/path`. Where a user is replicating from one S3 compatible service to another. Here we need a endpoint/ client per instance and will need to introduce a mechanism to alias one of the URIs so they are not identical.

 To do this without embedding endpoint info in the path then I think we need to allow the SPI to use some form of configuration file that optionally overloads default behavior.

Providing an alternative endpoint for scenario 2 should be possible by just using environment variables or any other standard way `S3ClientBuilder`s use to determine an endpoint.

For scenario 3 we will need a config file that has a mapping of bucket name to endpoint. Then when the SPI sees a path with a bucket name that matches the mapping it knows that those buckets should use the mapped endpoint preferentially to any other endpoint. If there is no mapping then it just lets the `S3ClientBuilder`s figure it out (or we could have a `*` mapping where any bucket not matching a specific mapping will use whatever client is mapped to `*`).

For scenario 4 we would need to provide a mapping of an alias to a real bucket name and for at least one of those buckets to have endpoint information. So scenario4 becomes `java MyApp s3://mybucket/path s3://mybucket-alias/path` and the config file lets the SPI know that when ever it sees `mybucket-alias` that it uses endpoint X and that the actual bucket name for `mybucket-alias` is actually `mybucket`.

stefanofornari:
I see the point that the s3:// URI is out of our control and may change in the future. Although we could handle a NIO provider URI as we want/need, I realize that if Amazon decides to change the scheme, this library and the clients of the library will be highly impacted (like lot of software around the world actually :)).
But ok, let's park the idea to overlap our provider URI scheme with AWS S3 scheme.

Having said that, IMHO the configuration approach would be quite hard to make work for us for two main reasons:

- credentials: orthogonal to the 4 cases there is that different endpoints (or even buckets) require each its own credentials
- being a library, we do not know in advance which endpoint/bucket will be accessed, which makes quite difficult to use a static configuration mechanism. plus we do not need to invent a new one, as FileSystems provides already newFileSystem(..., **env**)

we still have the problem of Paths.getPath(uri)...  However I would not reinvent the wheel, URI has already all the functionality we need.

Therefore I propose to walk another path you mentioned in an earlier message. What if we create another provider (still in this library), which defines a different URI scheme, let's say s3x:// . Here we can define how to structure the URI to address all the different cases and will be under our control. S3x:// will be used for 3rd party services only. We could have a S3XFileSystemProvider that inherits from S3FileSystemProvider and that turns the URI into the proper env to use in newFileSystem(). S3FileSystem will stay exactly the same. What do you think?

stefanofornari:
PS: deleted my branch by mistake.... :O it should be restored now...

markjschreiber:
> orthogonal to the 4 cases there is that different endpoints (or even buckets) require each its own credentials
Yikes! Yes that it true. It is easier if they are all Amazon S3 as you can assume a role which has access policies to both but definitely could be a major problem for different S3 compatible services.

> Therefore I propose to walk another path you mentioned in an earlier message. What if we create another provider (still in this library), which defines a different URI scheme, let's say s3x:// . Here we can define how to structure the URI to address all the different cases and will be under our control. S3x:// will be used for 3rd party services only. We could have a S3XFileSystemProvider that inherits from S3FileSystemProvider and that turns the URI into the proper env to use in newFileSystem(). S3FileSystem will stay exactly the same. What do you think?

I think this will probably be the best way to go as it will allow you to isolate a lot of specialization that you need but might not be compatible with Amazon S3's official `s3` scheme. Better yet, there is precedent for this. The Hadoop community uses the `s3a` scheme to refer to HDFS on S3. A different use case but the same idea of an S3 like URI that needs some special handling.

stefanofornari:
> I'm wondering if we can break this up into distinct pull requests?

I have created PR #114 to improve the configuration object S3NioSpiConfiguration with some suggestions @marco-luzzara. This is also required to better manage instantiation of S3FileSystems in S3FileSystemProvider's newFileSystem() (taking into account we are going to have two providers). See issue #112 for a better description.

stefanofornari:
> I'm wondering if we can break this up into distinct pull requests?

I would now move on with refactoring S3FileSystemProvider/S3FileSystem to be consistent with the NIO FileSystem behaviour. This includes:

- creating a new file system on newFileSystem() (i.e. no singleton); throw the proper exception if a FS for the same bucket has been already created
- returning a previously created FS on getFileSystem() or throw a proper exception if a FS for the bucket has not been created
- cache S3FileSystem instances (with a key derived from the bucket) instead of AWSClients
- use the new Configuration mechanism

We can refer to #30 for discussion.

@markjschreiber, let me know if you have any objections.

stefanofornari:
we can probably close this one too

Issue #112, opened 2023-07-15
Easy configuration to use in FileSystem.newFileSystem()


stefanofornari:
AS A user
I WANT to be able to provide configuration setting when creating a new file system. E.g.:
```

FileSystems.newFileSystem(myuri, Map.of(
    AWS_ACCESS_KEY_PROPERTY, "access-key",
    AWS_SECRET_ACCESS_KEY_PROPERTY, "secret-key",
    S3_SPI_ENDPOINT_PROTOCOL_PROPERTY, "http"
));
```

SO THAT I can easily provide configuration settings to the file system provider

**NOTES**

1. Use S3NioSpiConfigurationBuilder. E.g. :

```
FileSystems.newFileSystem(myuri, new S3NioSpiConfigurationBuilder()
    .withAccessKey("access-key")
    .withSecretKey("secret-key")
    .withEndpointProtocol("http")
);
```

2. for configuration parameter s3.spi.endpoint  no env var/sys prop overriding because not a good pattern with FileSystems where multiple instances of a file system can be created and are supposed to have their own env, while envvars and sysprops are global to the JVM.

markjschreiber:
Is it OK to close this? I think the mentioned PR includes this.

PR #114, opened 2023-07-15
Implementing fluent initialization/setting of S3NioSpiConfiguration

stefanofornari:
It also cleans up a bit the code using proper type for properties S3_SPI_READ_MAX_FRAGMENT_SIZE_PROPERTY and S3_SPI_READ_MAX_FRAGMENT_NUMBER_PROPERTY

*Issue #*
#112 

*Description of changes:*
See #112 for a description of the functionality introduced.

### DIFF
--- a/src/main/java/software/amazon/nio/spi/s3/config/S3NioSpiConfiguration.java
+++ b/src/main/java/software/amazon/nio/spi/s3/config/S3NioSpiConfiguration.java
@@ -240,15 +240,6 @@ public class S3NioSpiConfiguration extends HashMap<String, Object> {
      * @return this instance
      */
     public S3NioSpiConfiguration withCredentials(AwsCredentials credentials) {
-        //
-        // We do not store the credentials object directly for consistency with
-        // the other settings and to kee 1:1 relationship between the defined
-        // properties and the content of this map
-        //
-        if (credentials == null) {
-            return withCredentials(null, null);
-        }
-
         return withCredentials(credentials.accessKeyId(), credentials.secretAccessKey());
     }
 
--- a/src/main/java/software/amazon/nio/spi/s3/config/S3NioSpiConfiguration.java
+++ b/src/main/java/software/amazon/nio/spi/s3/config/S3NioSpiConfiguration.java
@@ -25,7 +25,7 @@ import software.amazon.nio.spi.s3.util.StringUtils;
 /**
  * Object to hold configuration of the S3 NIO SPI
  */
-public class S3NioSpiConfiguration extends HashMap<String, String> {
+public class S3NioSpiConfiguration extends HashMap<String, Object> {
 
     public static final String AWS_REGION_PROPERTY = "aws.region";
     public static final String AWS_ACCESS_KEY_PROPERTY = "aws.accessKey";
@@ -63,6 +63,11 @@ public class S3NioSpiConfiguration extends HashMap<String, String> {
      * The default value of the endpoint protocol property
      */
     public static final String S3_SPI_ENDPOINT_PROTOCOL_DEFAULT = "https";
+
+    /**
+     * The default value of the endpoint protocol property
+     */
+    public static final String S3_SPI_CREDENTIALS_PROPERTY = "s3.spi.credentials";
 
     private final Pattern ENDPOINT_REGEXP = Pattern.compile("(\\w[\\w\\-\\.]*)(?::(\\d+))");
     private final Logger logger = LoggerFactory.getLogger(this.getClass());
@@ -275,7 +280,7 @@ public class S3NioSpiConfiguration extends HashMap<String, String> {
      * @return the configured value or the default ("") if not overridden
      */
     public String getEndpoint() {
-        return getOrDefault(S3_SPI_ENDPOINT_PROPERTY, S3_SPI_ENDPOINT_DEFAULT);
+        return (String)getOrDefault(S3_SPI_ENDPOINT_PROPERTY, S3_SPI_ENDPOINT_DEFAULT);
     }
 
     /**
@@ -283,7 +288,7 @@ public class S3NioSpiConfiguration extends HashMap<String, String> {
      * @return the configured value or the default if not overridden
      */
     public String getEndpointProtocol() {
-        String protocol = getOrDefault(S3_SPI_ENDPOINT_PROTOCOL_PROPERTY, S3_SPI_ENDPOINT_PROTOCOL_DEFAULT);
+        String protocol = (String)getOrDefault(S3_SPI_ENDPOINT_PROTOCOL_PROPERTY, S3_SPI_ENDPOINT_PROTOCOL_DEFAULT);
         if ("http".equalsIgnoreCase(protocol) || "https".equalsIgnoreCase(protocol)) {
             return protocol;
         }
@@ -299,8 +304,8 @@ public class S3NioSpiConfiguration extends HashMap<String, String> {
     public AwsCredentials getCredentials() {
         if (containsKey(AWS_ACCESS_KEY_PROPERTY)) {
             return AwsBasicCredentials.create(
-                get(AWS_ACCESS_KEY_PROPERTY),
-                get(AWS_SECRET_ACCESS_KEY_PROPERTY)
+                (String)get(AWS_ACCESS_KEY_PROPERTY),
+                (String)get(AWS_SECRET_ACCESS_KEY_PROPERTY)
            );
         }
 
@@ -313,10 +318,8 @@ public class S3NioSpiConfiguration extends HashMap<String, String> {
      * @return the configured value or null if not provided
      */
     public String getRegion() {
-        return get(AWS_REGION_PROPERTY);
+        return (String)get(AWS_REGION_PROPERTY);
     }
-
-    // ------------------------------------------------------- protected methods
 
     /**
      * Generates an environment variable name from a property name. E.g 'some.property' becomes 'SOME_PROPERTY'
@@ -332,10 +335,8 @@ public class S3NioSpiConfiguration extends HashMap<String, String> {
                 .toUpperCase(Locale.ROOT);
     }
 
-    // --------------------------------------------------------- private methods
-
     private int parseIntProperty(String propName, int defaultVal){
-        String propertyVal = get(propName);
+        String propertyVal = (String)get(propName);
         try{
             return Integer.parseInt(propertyVal);
         } catch (NumberFormatException e){
--- a/src/main/java/software/amazon/nio/spi/s3/config/S3NioSpiConfiguration.java
+++ b/src/main/java/software/amazon/nio/spi/s3/config/S3NioSpiConfiguration.java
@@ -292,10 +292,9 @@ public class S3NioSpiConfiguration extends HashMap<String, Object> {
     /**
      * Get the configured credentials. Note that credentials can be provided in
      * two ways:
-     * <nl>
-     * <li>{@code withCredentials(String accessKey, String secretAcccessKey)}
-     * <li>{@code withCredentials(AwsCredentials credentials)}
-     * </nl>
+     * 
+     * 1. {@code withCredentials(String accessKey, String secretAcccessKey)}
+     * 2. {@code withCredentials(AwsCredentials credentials)}
      *
      * The latter takes the priority, so if both are used, {@code getCredentials()}
      * returns the most complete object, which is the value of the property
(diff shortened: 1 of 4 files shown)


src/main/java/software/amazon/nio/spi/s3/util/StringUtils.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package software.amazon.nio.spi.s3.util;
+
+import software.amazon.nio.spi.s3.S3Path;
+
+/**
+ *
+ */
+public class StringUtils {
+    public static String join(String separator, String[] elements, int startAt) {
```
markjschreiber, line 13: Possibly you can do this with a stream and collector or reuse `String.join()` with the sub array? https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#join-java.lang.CharSequence-java.lang.CharSequence...-

src/main/java/software/amazon/nio/spi/s3/util/StringUtils.java
```diff
@@ -0,0 +1,36 @@
...
+        StringBuilder sb = new StringBuilder();
+
+        for(; startAt < elements.length; ++startAt) {
+            sb.append(S3Path.PATH_SEPARATOR).append(elements[startAt]);
```
markjschreiber, line 17: do you mean `separator` rather than `S3Path.PATH_SEPARATOR`?

src/main/java/software/amazon/nio/spi/s3/util/StringUtils.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package software.amazon.nio.spi.s3.util;
+
+import software.amazon.nio.spi.s3.S3Path;
+
+/**
+ *
```
markjschreiber, line 10: Needs javadoc, including what happens to elements before the `startAt()`.

A more informative name might be useful as this is not just a normal `String.join(separator elements)`

src/main/java/software/amazon/nio/spi/s3/config/S3NioSpiConfiguration.java
```diff
@@ -27,86 +38,271 @@ public class S3NioSpiConfiguration {
...
+            put(AWS_REGION_PROPERTY, region.trim());
+        }
+
+        return this;
+    }
+
+    /**
+     * Fluently sets the value of accessKey and secretAccessKey
+     *
+     * @param accessKey the accesskey; if null, credentials are removed
```
markjschreiber, line 217: Could this method just take an instance of https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/auth/credentials/AwsCredentials.html

markjschreiber (changes requested):

src/main/java/software/amazon/nio/spi/s3/util/StringUtils.java
stefanofornari: Looking at the JDK code I found there is a StringJoin that is supposed to do help in joining strings :) I would not complicate more the implementation as being a utility method I want to keep the implementation as most efficient as possible.

src/main/java/software/amazon/nio/spi/s3/util/StringUtils.java
stefanofornari: Good catch.

src/main/java/software/amazon/nio/spi/s3/util/StringUtils.java
```diff
@@ -0,0 +1,36 @@
...
+     * contains only white space codepoints, otherwise false.
+     * 
+     * @param s the string to check
+     * 
+     * @return true if the string is empty or contains only white space codepoints, otherwise false
+     * 
+     * TODO: to replace with Java11's isBlank when moving away from Java8
+     */
+    public static boolean isBlank(final String s) {
+        return s.replace(" ", "").replace("\n", "").replace("\r", "").replace("\t", "").isEmpty();
```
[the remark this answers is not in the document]
stefanofornari: uhhh... mine was a pretty bad implementation actually... I am replacing with a much more efficient version (with a simple loop... less is more...)

src/main/java/software/amazon/nio/spi/s3/config/S3NioSpiConfiguration.java
```diff
@@ -27,86 +38,271 @@ public class S3NioSpiConfiguration {
...
-        this(new Properties());
+        overrides.keySet().forEach(key -> put(key, String.valueOf(overrides.get(key))));
     }
 
     /**
      * Create a new configuration with overrides
      * @param overrides the overrides
      */
+    //
+    // TODO: to be removed if not used
```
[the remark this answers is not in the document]
stefanofornari: yeah ... as you probably noticed, I tend to I value clean code more than stability of the interface :) but in this case I agree being a visible interface.

src/main/java/software/amazon/nio/spi/s3/config/S3NioSpiConfiguration.java
```diff
@@ -27,86 +38,271 @@ public class S3NioSpiConfiguration {
...
+    }
+
+    /**
+     * Fluently sets the value of the endpoint's protocol
+     *
+     * @param protocol the endpoint's protcol
+     *
+     * @return this instance
+     */
+    public S3NioSpiConfiguration withEndpointProtocol(String protocol) {
```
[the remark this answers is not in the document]
stefanofornari: Good question. I thought about it, the problem is as usual Paths.getPath("s3x://endpoint/bucket"). Supporting Paths.getPath("s3x://htpps://endpoint/bucket") is not an option. I was thinking of splitting the scheme in s3x and s3xs for https ... I do not like it that much. Another option could be to use a query string parameter.
Anyway, for wow I need a way to test on a HTTP endpoint, therefore I would leave it as is for the moment.

src/main/java/software/amazon/nio/spi/s3/config/S3NioSpiConfiguration.java
```diff
@@ -27,86 +38,271 @@ public class S3NioSpiConfiguration {
...
+     * The name of the endpoint protocol property
+     */
+    public static final String S3_SPI_ENDPOINT_PROTOCOL_PROPERTY = "s3.spi.endpoint-protocol";
+    /**
+     * The default value of the endpoint protocol property
+     */
+    public static final String S3_SPI_ENDPOINT_PROTOCOL_DEFAULT = "https";
 
-    private final Properties properties;
+    private final Pattern ENDPOINT_REGEXP = Pattern.compile("(\\w[\\w\\-\\.]*)(?::(\\d+))");
```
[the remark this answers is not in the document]
stefanofornari: Yes, IP addresses must be supported (see also #115)

src/main/java/software/amazon/nio/spi/s3/config/S3NioSpiConfiguration.java
stefanofornari: Yep, that could be an option; I actually consider it, but look at the required code from a user perspective:

option1: config.withCredentials("key", "secret");
option2: config.withCredentials(AwsBasicCredentials.create("key", "secret")); (plus the import)

I would prefer option1 :) - although we can have an additional withCredentials(cred) in the case cred is already a AwsCredentials().

markjschreiber: Using access key and secret access key are but one way to create credentials in AWS. There are also situations in which you would need to provide additional information such as when using temporary credentials. The AWS docs list the full range of environment variables which can be used. I would therefore remove this version of `withCredentials` and use only `withCredentials(AwsCredentials)` and store that object in the map if needed.


If desired, you could retain this version of the method as a convenience method but use it to build an `AwsBasicCredentials` object and then use that in `withCredentials(credentials). An inversion of what you have here.

src/main/java/software/amazon/nio/spi/s3/config/S3NioSpiConfiguration.java
```diff
@@ -75,38 +115,209 @@ public S3NioSpiConfiguration(){
...
+    /**
+     * Get the configured region if any
+     *
+     * @return the configured value or null if not provided
+     */
+    public String getRegion() {
+        return get(AWS_REGION_PROPERTY);
     }
 
+    // ------------------------------------------------------- protected methods
```
markjschreiber, line 319: remove this comment, we haven't used this convention in other places so probably shouldn't introduce it here

src/main/java/software/amazon/nio/spi/s3/config/S3NioSpiConfiguration.java
```diff
@@ -120,4 +331,17 @@ protected String convertPropertyNameToEnvVar(String propertyName){
                 .replace('.', '_').replace('-', '_')
                 .toUpperCase(Locale.ROOT);
     }
+
+    // --------------------------------------------------------- private methods
```
markjschreiber, line 335: remove comment

src/test/java/software/amazon/nio/spi/s3/config/S3NioSpiConfigurationTest.java
```diff
@@ -36,24 +43,164 @@ public void setup(){
...
+    @Test
+    public void withAndGetEndpoint() {
+        then(config.withEndpoint("somewhere.com:8000")).isSameAs(config);
+        then(config.getEndpoint()).isEqualTo("somewhere.com:8000");
+        then(config.withEndpoint(" somewhere.com:8080\t").getEndpoint()).isEqualTo("somewhere.com:8080");
+        then(config.withEndpoint("   ").getEndpoint()).isEqualTo("");
+        then(config.withEndpoint(null).getEndpoint()).isEqualTo("");
+
+        try {
+           config.withEndpoint("noport.somewhere.com");
```
markjschreiber, line 106: This pattern should be allowed. Port numbers are not usually present in official AWS service endpoints which look like `https://<service>.us-east-1.amazonaws.com`

markjschreiber (changes requested):
I think we are getting there. A couple of changes still required.

stefanofornari:
I should have resolved all comments, but see below.  There is an odd issue with gradle... I do not know why it is complaining about the version of a class in the logback package....
```
java.lang.UnsupportedClassVersionError: ch/qos/logback/classic/spi/LogbackServiceProvider has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
[...]
```
I do not have the problem with maven and  I am not a gradle expert. Can you give me some help?

> Using access key and secret access key are but one way to create credentials in AWS. There are also situations in which you would need to provide additional information such as when using temporary credentials. The AWS docs list the full range of environment variables which can be used. I would therefore remove this version of withCredentials and use only withCredentials(AwsCredentials) and store that object in the map if needed.
> 
> If desired, you could retain this version of the method as a convenience method but use it to build an AwsBasicCredentials object and then use that in `withCredentials(credentials). An inversion of what you have here.

I have implemented it as suggested for your review. However I am doubtful: this configuration class is mostly designed to glue  envvar/syspro and simple configuration properties. WRT credentials the original implementation did not provide special support credentials because it delegated it to the underlying AWS library. To support 3rd party services we need instead to add a key/secret mechanism. The doubt is that support for different authentication means, usually AWS specific, can still be achieved without the need of withCredentials(AwsCredentials) just not setting any credentials.
In short: consider to get rid of it, I am fine either ways.

stefanofornari:
@markjschreiber any chances we can sort out this one? it would be very helpful to start to work on s3x provider

markjschreiber:
> @markjschreiber any chances we can sort out this one? it would be very helpful to start to work on s3x provider

Almost. I think that if you want to retain the `withCredentials(string, string)` method then this should construct an `AwsBasicCredentials` instance and pass that the `withCredentials(AwsCredentials)` and not the other way around. That way people with more complex credentials can use the `withCredentials(AwsCredentials)` version.

Apart from that it looks good although one of the unit tests seems to be failing right now.

src/test/java/software/amazon/nio/spi/s3/config/S3NioSpiConfigurationTest.java
stefanofornari: done

src/main/java/software/amazon/nio/spi/s3/config/S3NioSpiConfiguration.java
stefanofornari: done

stefanofornari:
@markjschreiber it should be ready to go.

markjschreiber (approved):
Looks good. Thanks for the PR!

PR #168, opened 2023-09-10
Third party S3 services support + #61

### LINKED ISSUE #61: File System settings

stefanofornari:
This pull request should finalize the work to seamlessly support 3rd party S3 services like minio. It brings three main improvements, which I kept together as they are small enough and they complement each other.

1. All information to build a file system (bucket, endpoint, credentials, etc.) are now encapsulated in a S3NioConfiguration object that is passed to the constructor; this simplifies quite a bit the constructors and the need of accessors. This should also close #61 
2. Deprecation of constructors taking a URI: in the NIO world, the FileSystem object is usually create indirectly, by the provider or even more subtly by using Files or Path; it is therefore cleaner the provider digests the URIs creating a corresponding configuration object that can then be passed to S3FileSystem's constructor. This brings also the valuable advantage of making the file S3FileSystem completely independent by the URI used to create and configure it.
3. Last but most importantly, it introduces a new file system provider, which manages the s3x:// URI schema as per the following URI pattern (as discussed earlier in other tickets and PRs):

    `s3x://[key:secret#@]hostename[:port]/bucket/key`

Thanks to bullet 1 above, it handed up to be quite clean, as the new S3XFileSystemProvider extends S3FileSystemProvider with very little code, while S3FileSystem is i completely in common between the two providers.

@markjschreiber , let me know what you think

### DIFF
--- a/src/main/java/software/amazon/nio/spi/s3/S3ClientProvider.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3ClientProvider.java
@@ -35,6 +35,7 @@ import software.amazon.awssdk.services.s3.S3CrtAsyncClientBuilder;
 import software.amazon.awssdk.services.s3.model.HeadBucketResponse;
 import software.amazon.awssdk.services.s3.model.S3Exception;
 import software.amazon.nio.spi.s3.config.S3NioSpiConfiguration;
+import static software.amazon.nio.spi.s3.util.StringUtils.isBlank;
 
 /**
  * Factory/builder class that creates sync and async S3 clients. It also provides
@@ -70,9 +71,9 @@ public class S3ClientProvider {
             .build();
 
     private final EqualJitterBackoffStrategy backoffStrategy = EqualJitterBackoffStrategy.builder()
-        .baseDelay(Duration.ofMillis(200L))
-        .maxBackoffTime(Duration.ofSeconds(5L))
-        .build();
+            .baseDelay(Duration.ofMillis(200L))
+            .maxBackoffTime(Duration.ofSeconds(5L))
+            .build();
 
     final RetryCondition retryCondition;
 
@@ -106,6 +107,14 @@ public class S3ClientProvider {
 
     public S3ClientProvider() {
         this(null);
+    }
+
+    public S3CrtAsyncClientBuilder asyncClientBuilder() {
+        return asyncClientBuilder;
+    }
+
+    public void asyncClientBuilder(final S3CrtAsyncClientBuilder builder) {
+        asyncClientBuilder = builder;
     }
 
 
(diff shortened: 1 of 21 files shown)


src/main/java/software/amazon/nio/spi/s3/S3ClientProvider.java
```diff
@@ -126,32 +149,41 @@ protected S3Client generateClient(String bucketName){
...
     protected S3Client generateClient (String bucketName, S3Client locationClient) {
         logger.debug("generating client for bucket: '{}'", bucketName);
         S3Client bucketSpecificClient = null;
 
+        if ((configuration.getEndpoint() == null) || isBlank(configuration.getEndpoint())) {
+            //
+            // we try to locate a bucket only if no endpoint is provided, which
+            // means we are dealing with AWS S3 buckets
+            //
         String bucketLocation = null;
```
markjschreiber, line 187: Nit. The content of the if block should be indented

src/main/java/software/amazon/nio/spi/s3/S3ClientProvider.java
```diff
@@ -186,29 +218,32 @@ protected S3Client generateClient (String bucketName, S3Client locationClient) {
...
     protected S3AsyncClient generateAsyncClient (String bucketName, S3Client locationClient) {
         logger.debug("generating asynchronous client for bucket: '{}'", bucketName);
         S3AsyncClient bucketSpecificClient = null;
 
+        if ((configuration.getEndpoint() == null) || isBlank(configuration.getEndpoint())) {
+            //
+            // we try to locate a bucket only if no endpoint is provided, which
+            // means we are dealing with AWS S3 buckets
+            //
         String bucketLocation = null;
```
markjschreiber, line 247: Indent the content of the `if` block

markjschreiber (commented):
Looks great. There is one minor comment about some `if` statements that appear to not be indented but I think we are looking good.

stefanofornari:
fixed, thanks!

markjschreiber (approved):
Thanks for this contribution and for adding support for alternative S3 providers. I will bump the minor version later this week and make a release to maven central.

stefanofornari:
@markjschreiber do we have the new version on central? 
Also, I have created a couple of minor "cleanup" PRs. Would you mind review and merge them as well?

Issue #161, opened 2023-09-06
S3NioSpiConfiguration constructor with overrides converts items to Strings


stefanofornari:
See https://github.com/stefanofornari/aws-java-nio-spi-for-s3/issues/3

stefanofornari:
fixed in v1.2.4

Issue #115, opened 2023-07-15
Bug in S3Client?


stefanofornari:
Hi, sorry to spam this repository, but can you point me where I can file the following bug.

**Steps to reproduce**

1. set up a fake listener (e,g, nc -l 8880)
2. Make a list objects call:
```
S3Client s3Client = S3Client.builder()
    .endpointOverride(URI.create("http://localhost:8880"))
    .region(Region.EU_CENTRAL_1)
    .credentialsProvider(() -> AwsBasicCredentials.create("key", "secret"))
    .build();
    s3Client.listObjectsV2(
        ListObjectsV2Request.builder().bucket("pippo").build()
    );
```

**Expected behaviour**
A request to /pippo is issued, like:
```
GET /pippo?list-type=2 HTTP/1.1
Host: localhost:8880
amz-sdk-invocation-id: cef3a91f-bf57-cfbb-1fb8-45abea064b71
amz-sdk-request: attempt=1; max=4
Authorization: AWS4-HMAC-SHA256 Credential=key/20230715/eu-central-1/s3/aws4_request, SignedHeaders=amz-sdk-invocation-id;amz-sdk-request;host;x-amz-content-sha256;x-amz-date, Signature=ce93b4debefcf8f94fc57a6107584b06cbe712fe4d8228f77a69599b04997be2
User-Agent: aws-sdk-java/2.20.79 Linux/5.4.0-150-generic OpenJDK_64-Bit_Server_VM/19.0.1+10 Java/19.0.1 vendor/Azul_Systems__Inc. io/sync http/Apache cfg/retry-mode/legacy ft/s3-transfer
x-amz-content-sha256: UNSIGNED-PAYLOAD
X-Amz-Date: 20230715T093511Z
Connection: Keep-Alive
```
**Current result**
```
GET /?list-type=2 HTTP/1.1
Host: pippo.localhost:8880
amz-sdk-invocation-id: e6dd0685-09a1-807e-c9a0-7ead5f096fcb
amz-sdk-request: attempt=1; max=4
Authorization: AWS4-HMAC-SHA256 Credential=key/20230715/eu-central-1/s3/aws4_request, SignedHeaders=amz-sdk-invocation-id;amz-sdk-request;host;x-amz-content-sha256;x-amz-date, Signature=543f22241699c0337c2a17b9d07c84620255a371098dc636ac78ff0f351a18bf
[...]

PR #174, opened 2023-09-14
Fixing issues #115 and #161

stefanofornari:
*Issue #, if available:*
This PR covers issue #115 and #161 

S3 compatible services like minio do not like AWS S3 virtual-style addressing, where the bucket name is considered part of the hostname (e.g. pippo -> pippo.s3.eu-central-1.amazonaws.com).
We therefore set the path-style addressing by default for s3x provider, while we also give the possibility to configure it via the s3.spi.force-path-style.

I took this opportunity to make some improvements to the readme about the different configuration parameters. Please have a look and let me know if you like it. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,11 +122,34 @@ on I/O, up to the limits of your network connection.
 
 ### Configuration
 
-The read-ahead buffer asynchronously prefetches `n` sequential fragments of `m` bytes from S3. The
-values of `n` and `m` can be configured to your needs by using command line properties or environment variables.
+There are two types of configuration parameters: local and system. Local configuration parameters are set when creating
+a file system calling _FileSystems.newFileSystem(uri, env)_ or directly calling _newFileSystem(uri, env)_ on a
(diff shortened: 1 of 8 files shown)


markjschreiber:
@stefanofornari there is a version conflict in this PR. This may be due to the order in which I merged your other PR? Can you resolve the conflict and then we can get these changes into the next release.

stefanofornari:
> @stefanofornari there is a version conflict in this PR. This may be due to the order in which I merged your other PR? Can you resolve the conflict and then we can get these changes into the next release.

I should have done

markjschreiber (commented):
I like this PR but I am wondering if there is a way to only force path style for `S3x` URIs? Path style requests are being retired from S3. Could the default be to not force path style but to allow that to be configured?

https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#path-style-access

markjschreiber (approved):
Please ignore my last comment. I missed that force path style isn't actually the default.

PR #192, opened 2023-09-29
chore(deps): Remove duplicate `rxjava` dependency

guicamest:
This PR removes a duplicate declaration of the dependency on rxjava


markjschreiber (approved):
Thanks for fixing this up

markjschreiber:
I ended up combining this with the dependabot version bump for rxjava so as to avoid you having to rebase this.

PR #195, opened 2023-09-29
chore(test): Improve S3FileSystemProviderTest robustness

guicamest:
*Description of changes:*

After checking out the project and running the tests, they would fail or not depending on the order the tests are run (it is not guaranteed they are by default).

I've noticed `S3PathTest.testEquals` creates `foo` _fs_ but it is not closed; `S3FileSystemProviderTest` uses `foo` _fs_ for its tests.
So this PR introduces 2 changes:
* Cleanup _fs_ in `testEquals` after creating it
* Get or create _fs_ in `S3FileSystemProviderTest` so that it is not affected by other tests leaking resources


markjschreiber (approved):
Thanks for spotting and fixing this problem!

PR #199, opened 2023-10-02
chore(test): Replace try/fail/catch/assert with assertThat

guicamest:
*Description of changes:*
This PR replaces several occurrences of the idiom:
```
try {
  ... // do something that throws an Exception
  fail() // with some message to display because the exception should have been thrown
} catch ( someException ){
  ... // some assertion on the exception
}
```
with a couple of assertion methods provided by AssertJ (`assertThatCode`, `assertThatThrownBy`) and JUnit (`assertThrows`) which are already part of the dependencies of the project.

Before:
```
        try {
            provider.newFileSystem(URI.create(pathUri));
            fail("filesystem created twice!");
        } catch (FileSystemAlreadyExistsException x) {
            assertTrue(x.getMessage().contains("'foo'"));
        }
```
After:
```
assertThatCode(() -> provider.newFileSystem(URI.create(pathUri)))
                .as("filesystem created twice!")
                .isInstanceOf(FileSystemAlreadyExistsException.class)
                .hasMessageContaining("'foo'");
```

### DIFF
--- a/src/test/java/software/amazon/nio/spi/s3/S3ClientProviderTest.java
+++ b/src/test/java/software/amazon/nio/spi/s3/S3ClientProviderTest.java
@@ -186,11 +186,7 @@ public class S3ClientProviderTest {
         );
 
         // then you should get a NoSuchElement exception when you try to get the header
-        try {
-            provider.generateAsyncClient("test-bucket", mockClient);
-        } catch (Exception e) {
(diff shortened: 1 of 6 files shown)


markjschreiber (approved):
Nice refactoring to assertJ idioms. Thanks!

Issue #154, opened 2023-08-30
logback.xml file interfere with project's default config


aliabdikk:
I was trying to use your library in my Micronaut project that already has Logback configured.
However, since your library is also featuring a Logback,.xml config file, it interferes with my main config and messes up with the Slf4J logging.

`WARN in ch.qos.logback.classic.util.DefaultJoranConfigurator@4ef782af - Resource [logback.xml] occurs multiple times on the classpath.`

Would that be possible to make the Logback optional in your library? Usually libraries refrain from including Log configs so that it could be up to the end user to define them.

Thanks

PR #200, opened 2023-10-02
chore(logback): Fix `logback.xml file interfere with project's default config`

guicamest:
*Issue #154*

*Description of changes:*
This PR moves the `logback.xml` from `src/main` into `src/test` so that:
* Fixes the interference issue 
* Logs in tests are formatted as defined and log levels can be configured


markjschreiber (approved):
Thanks for this PR

PR #201, opened 2023-10-02
chore(S3FS): Extract supported file attribute views to static field

guicamest:
*Description of changes:*

This PR extracts `supportedFileAttributeViews` into a static field so that the creation of the temporary `HashSet` and the call to `unmodifiable...` do not happen each time the method is called. It also uses just the supported value instead of having a size of `2`.


markjschreiber (approved):
Approved. Thanks

Issue #194, opened 2023-09-29
S3WritableByteChannel does not consider contentType while putting an object into the bucket


aliabdikk:
Upon uploading a file to the bucket, we are not defining any content type:

https://github.com/awslabs/aws-java-nio-spi-for-s3/blob/5e821745263c57d7c00df397a8ce425d3a01a446/src/main/java/software/amazon/nio/spi/s3/S3WritableByteChannel.java#L102-L128

This could be crucial, like when we are uploading a special file (like XML file) to be used by another application later.

I believe something like `.contentType(Files.probeContentType(tempFile))` before building the `putObjectRequest` should do the trick.

markjschreiber:
This is a good suggestion. Looking at the JDK docs I can see that `Files.probeContentType(tempFile)` will return `null` if the contentType cannot be guessed so we will need to check that the `putObjectRequest` will be OK with the value of `.contentType()` being null.

It may take me a couple of weeks to get to this, but feel free to make a PR if you want to see this added faster.

aliabdikk:
@markjschreiber Done:
https://github.com/awslabs/aws-java-nio-spi-for-s3/pull/204

markjschreiber:
PR Merged

PR #204, opened 2023-10-04
Provide content-type while putting object to S3

aliabdikk:
*Issue #194:*
[S3WritableByteChannel does not consider contentType while putting an object into the bucket](https://github.com/awslabs/aws-java-nio-spi-for-s3/issues/194)

*Description of changes:*
Upon putting an object into the S3 bucket (for instance, while copying), we need to also provide the content-type of the file. This could come crucial in cases like when the uploaded file (let's say an XML file) is going to be used by a specific application later.
While checking the implementation of `PutObjectRequest`, I didn't fine anywhere that `contentType` acquires a default value, so if `Files::probeContentType` returns a null value, it is the same as not calling the builder method. Hence I don't see any negative side-effects with this change.


markjschreiber (approved):
Thanks for this

PR #211, opened 2023-10-05
chore(README): Format local/system parameters as table

guicamest:
*Description of changes:*

This PR formats local and system parameters in the README as a table, in order to improve reading


guicamest:
I think it can even be written as

|parameter|type|description|
|---------|-----|------|
|**aws.region**|local/system|specifies the default region for API calls|
|..|..|
|**s3.spi.endpoint**|local|the endpoint to use to access the bucket; this is extracted from the uri by the S3X provider|
|..|..|

What do you think?

PR #225, opened 2023-10-24
chore: extract method to check Path is S3Path

guicamest:
*Description of changes:*

Considering that in many places a check is done to verify the `path` (from the nio public API signature) is an `S3Path`, the changes introduce a method `checkPath` to avoid repetition of the code, better reusability and consistency (checks were done differently depending on the method).

The implementation is quite similar to the one used in `UnixFileSystemProvider` which is quite straight forward, replacing the path implementation with `S3Path`:
```
    UnixPath checkPath(Path obj) {
        if (obj == null)
            throw new NullPointerException();
        if (!(obj instanceof UnixPath))
            throw new ProviderMismatchException();
        return (UnixPath)obj;
    }
```

### DIFF
--- a/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
@@ -485,7 +485,7 @@ public class S3FileSystemProvider extends FileSystemProvider {
         S3Path s3Directory = checkPath(dir);
         S3FileSystem fs = s3Directory.getFileSystem();
         try {
-            String directoryKey = (s3Directory).toRealPath(NOFOLLOW_LINKS).getKey();
+            String directoryKey = s3Directory.toRealPath(NOFOLLOW_LINKS).getKey();
             if (!directoryKey.endsWith(S3Path.PATH_SEPARATOR) && !directoryKey.isEmpty()) {
                 directoryKey = directoryKey + S3Path.PATH_SEPARATOR;
             }
@@ -1135,8 +1135,7 @@ public class S3FileSystemProvider extends FileSystemProvider {
     }
 
     protected static S3Path checkPath(Path obj) {
-        if (obj == null)
-            throw new NullPointerException();
+        Objects.requireNonNull(obj);
         if (!(obj instanceof S3Path))
             throw new ProviderMismatchException();
         return (S3Path)obj;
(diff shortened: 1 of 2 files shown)


src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
```diff
@@ -1142,4 +1133,12 @@ private S3Path forceAwsClient(final Path path, final S3AsyncClient client) {
         p.getFileSystem().clientProvider(new FixedS3ClientProvider(client));
         return p;
     }
+
+    protected static S3Path checkPath(Path obj) {
+        if (obj == null)
```
markjschreiber, line 1138: You could use `Objects.requireNotNull(obj)`

src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
```diff
@@ -482,9 +482,10 @@ private Iterator<Path> pathIteratorForPublisher (
      */
     @Override
     public void createDirectory(Path dir, FileAttribute<?>... attrs) throws IOException {
-        S3FileSystem fs = ((S3Path)dir).getFileSystem();
+        S3Path s3Directory = checkPath(dir);
+        S3FileSystem fs = s3Directory.getFileSystem();
         try {
-            String directoryKey = ((S3Path)dir).toRealPath(NOFOLLOW_LINKS).getKey();
+            String directoryKey = (s3Directory).toRealPath(NOFOLLOW_LINKS).getKey();
```
markjschreiber, line 488: remove enclosing braces?

markjschreiber (changes requested):
Thanks for this PR. A few minor changes requested.

src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
guicamest: Changed in 62b7f74

src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
guicamest: Changed in 62b7f74

markjschreiber (approved):

markjschreiber:
Thanks again. Approved and merged!

PR #227, opened 2023-10-26
refactor: Reduce API implementation classes visibility

guicamest:
## Description of changes

As described in https://github.com/awslabs/aws-java-nio-spi-for-s3/pull/223#issuecomment-1778739120:
>  Classes visibility are reduced in order to avoid leaking the classes to the user of the library who could potentially write some code with this information. This will benefit the maintenance of the library, as it'll provide more freedom to implementation changes without risking breaking user code. After all, **this library is an implementation of NIO** so that users shouldn't care if a Path is an `S3Path` or `AWSS3Path`. This is how the default implementations are implemented, for example, `class LinuxFileSystem` is not public.

### Code coverage
image

image


guicamest:
Looking at [FileSystems.newFileSystem](https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileSystems.html#newFileSystem-java.net.URI-java.util.Map-) javadoc, it looks like the method is intented for creating filesystems, not just a `FileSystem` object. If this is correct, then the implementation of `newFileSystem` in the `S3FileSystemProvider` would be incorrect, as it is getting a `FileSystem` object, not *creating* a `FileSystem`. Given the description on the README about the relationship `FileSystem` *<->* `Buckets`, it would mean *creating* a Bucket).

The default FileSystem implementations `throw` when `newFileSystem` is called:
image
image

IMHO, I think it will be a good opportunity to remove the current implementation of newFileSystem for _v2.x.x_ , leaving it unimplemented (`UnsupportedOperation`) until:
* it is implemented for bucket creation (new feature, non breaking) or,
* left unimplemented but following the specification

What are your thoughts @markjschreiber ?

markjschreiber:
You make a good point. I would be in favor of being able to use it to create a bucket. The map of key value pairs passed to it would need to map to properties that can be set on the `CreateBucketRequest`.

guicamest:
>  We would need to move the current functionality of the newFileSystem over to the getFileSystem method on the S3FileSystemProvider.

Exactly!

Would it be ok to turn it into unimplemented for this PR and leave the implementation for a future PR? Otherwise it'll be a hell to review as it is already big.

src/main/java/software/amazon/nio/spi/s3/Constants.java
```diff
@@ -0,0 +1,5 @@
+package software.amazon.nio.spi.s3;
+
+public class Constants {
```
markjschreiber, line 3: I wonder if there is any value in also placing the scheme `s3` here as well?

src/main/java/software/amazon/nio/spi/s3/NotYetImplementedException.java
```diff
@@ -0,0 +1,7 @@
+package software.amazon.nio.spi.s3;
+
+public class NotYetImplementedException extends RuntimeException {
```
markjschreiber, line 3: Why not just use the Java `UnsupportedOperationException`?

src/main/java/software/amazon/nio/spi/s3/S3FileSystem.java
```diff
@@ -420,19 +371,66 @@ public WatchService newWatchService() {
...
+     */
+    String bucketName() {
+        return bucketName;
+    }
+
     /**
      * The list of currently open channels. Exposed mainly for testing
      * @return a read only view wrapping the set of currently open channels.
      */
-    protected Set<Channel> getOpenChannels(){
```
markjschreiber, line 427: This actually increases visibility to the package level. That is fine if that is what you intended.

src/main/java/software/amazon/nio/spi/s3/S3FileSystem.java
```diff
@@ -420,19 +371,66 @@ public WatchService newWatchService() {
...
+    Set<Channel> getOpenChannels(){
         return Collections.unmodifiableSet(openChannels);
     }
 
-    protected void registerOpenChannel(S3SeekableByteChannel channel){
```
markjschreiber, line 431: This actually increases visibility to the package level. That is fine if that is what you intended.

src/main/java/software/amazon/nio/spi/s3/S3FileSystem.java
```diff
@@ -420,19 +371,66 @@ public WatchService newWatchService() {
...
+    void registerOpenChannel(S3SeekableByteChannel channel){
         openChannels.add(channel);
     }
 
-    protected boolean deregisterClosedChannel(S3SeekableByteChannel closedChannel){
```
markjschreiber, line 435: This actually increases visibility to the package level. That is fine if that is what you intended.

src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
```diff
@@ -74,7 +75,7 @@ public class S3FileSystemProvider extends FileSystemProvider {
     /**
      * Constant for the S3 scheme "s3"
      */
-    public static final String SCHEME = "s3";
+    static final String SCHEME = "s3";
```
markjschreiber, line 78: Consider moving to `Constants`

src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
```diff
@@ -90,88 +91,12 @@ public String getScheme() {
...
-     *
-     * @throws FileSystemAlreadyExistsException if the file system has already been created
-     * @throws IllegalArgumentException if the pre-conditions for the uri parameter
-     *         are not met, or the env parameter does not contain properties
-     *         required by the provider, or a property value is invalid
-     */
-    public S3FileSystem newFileSystem(URI uri) {
-        return newFileSystem(uri, Collections.EMPTY_MAP);
+    public FileSystem newFileSystem(URI uri, Map<String, ?> env) {
+        throw new NotYetImplementedException("This method is not yet supported in v2.x. It might be implemented for bucket creation");
```
markjschreiber, line 99: Can you add an issue to implement this for bucket creation?

src/main/java/software/amazon/nio/spi/s3/S3XFileSystemProvider.java
```diff
@@ -14,34 +14,16 @@
  * limitations under the License.
  */
 
-package software.amazon.nio.spi.s3x;
+package software.amazon.nio.spi.s3;
```
markjschreiber, line 17: You seem to have moved some of the s3x... classes into the s3 package. Is this intentional? If so you should probably move the rest as well?

markjschreiber (changes requested):
Thanks for this refactoring. I have asked a couple of questions in the review and there are some suggested changes.

src/main/java/software/amazon/nio/spi/s3/NotYetImplementedException.java
guicamest: Good point. That was my original approach, but by doing so, `FileSystems.newFileSystem(uri, Collections.emptyMap())` fails with
> java.nio.file.ProviderNotFoundException: Provider "s3" not found

which is quite a _misleading_ message. When reading the message, it can make one doubt that the provider is **actually** present (and correctly working) or not :( 

image

src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
guicamest: https://github.com/awslabs/aws-java-nio-spi-for-s3/pull/227#discussion_r1376624095

src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
guicamest: Sure I'll do

src/main/java/software/amazon/nio/spi/s3/S3XFileSystemProvider.java
guicamest: Yes it was intentional. By changing the visibility of `fileSystemInfo` in the superclass of the `S3XFileSystemProvider`, I needed to move this one to the same package.

The one left in `s3x` package is `S3XFileSystemInfo`, I can move it👍

markjschreiber:
Merging the other PRs has created a conflict here. You can probably resolve it by rebasing this onto HEAD

src/main/java/software/amazon/nio/spi/s3/S3XFileSystemProvider.java
guicamest: Done in 1c80df9

src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
guicamest: Created https://github.com/awslabs/aws-java-nio-spi-for-s3/issues/243, feel free to add any information you consider relevant 🙏

src/main/java/software/amazon/nio/spi/s3/S3FileSystem.java
guicamest: `protected` [already allows](https://www.baeldung.com/java-protected-access-modifier#1-from-the-same-package) _same package_ access 😅 

The change actually helps to hide `S3SeekableByteChannel`:

image
--
image

markjschreiber (approved):
Looks great

PR #248, opened 2023-10-31
refactor(S3ClientProvider): Reduce complexity of `generateClient`

guicamest:
*Description of changes:*

The `generateClient` (and `generateAsyncClient`) method was quite complex, had several nested blocks (if/try/if/try/if). Some new methods have been extracted in order to make the code more readable and maintainable.

Both `generateClient` and `generateAsyncClient` shared the same logic - except for the underlying client that was generated. A new method `getClientForBucket` has been extracted that receives a function to construct the client for the given bucket location. The _old_ `generateClient` has been renamed to `generateSyncClient`.


markjschreiber (approved):
Thanks for tidying up this code

PR #256, opened 2023-11-02
log(S3FileSystem): Use scheme of the provider instead of `s3`

guicamest:
*Description of changes:*

Since there are two schemes `s3` and `s3x`, it is convenient that the logs reflect the proper scheme.


markjschreiber (approved):
Good catch!

PR #258, opened 2023-11-03
rewrite(S3SeekableByteChannel): Use `S3BasicFileAttributes` to get "file"size of path

guicamest:
*Description of changes:*

The call to get the "file" size in `S3SeekableByteChannel` is replaced with `S3BasicFileAttributes.size()` since it already provides a way to get a "file" size.

_Package Coverage_ (Before / After)
Screenshot 2023-11-03 095239

_Class Coverage_ (Before / After)
Screenshot 2023-11-03 095256


markjschreiber (approved):
Nice cleanup of repeated code.

PR #286, opened 2023-11-08
test(S3BasicFileAttributes): When a timeout happens getting attributes of a regular file, an Exception should be thrown

guicamest:
*Description of changes:*

The PR introduces a test for `S3BasicFileAttributes`, _when a timeout happens getting attributes of a regular file, an Exception should be thrown_. In order to achieve this, and avoid having to wait 1 minute( the _current_ default ) for the test to fail, the timeout is now configurable.

Coverage:
Screenshot 2023-11-08 130844

By the way, isn't 1 minute to high for a timeout? I imagine a user waiting up to 1 minute to get the size of a file would be too much and a bad user experience. I was thinking maybe 10 seconds would be a more sensible default. What are your thoughts @markjschreiber ?

### DIFF
--- a/src/main/java/software/amazon/nio/spi/s3/S3BasicFileAttributeView.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3BasicFileAttributeView.java
@@ -5,9 +5,12 @@
 
 package software.amazon.nio.spi.s3;
 
+import software.amazon.nio.spi.s3.util.TimeOutUtils;
+
 import java.nio.file.attribute.BasicFileAttributeView;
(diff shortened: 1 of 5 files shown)


markjschreiber (approved):
Looks great. Thanks for the contribution.

guicamest:
Happy to contribute! What do you think about reducing the timeout for the attributes @markjschreiber ?
> By the way, isn't 1 minute to high for a timeout? I imagine a user waiting up to 1 minute to get the size of a file would be too much and a bad user experience. I was thinking maybe 10 seconds would be a more sensible default.

PR #287, opened 2023-11-08
rewrite: Use type inference when possible

guicamest:
*Description of changes:*

Java 10 introduces `var` which uses type inference and saves the developer from specifying the type of a variable. This also helps to remove some imports which are only used to declare the type of a variable.


markjschreiber (approved):
Nice refactor. Thanks

stefanofornari:
I loved to have discussed this change. Use of var in Java is quite arguable and there are good reasons not to use it. It is actually pretty bad practice IMHO (and not only) which lower the quality of the code with no real advantage.

markjschreiber:
I know there is active debate about the use of `var` in Java and I didn't originally like it. However, there are many strictly typed languages that allow compile time type inference including Scala, Kotlin and even Rust. I'm willing to defer to the designers of these languages (and Java) that the pros probably out weigh the cons.

Dynamic inference at runtime is a terrible idea though, I hope Java never goes there!

stefanofornari:
I agree to let the debate about introducing var in Java to who designs the language. I even appreciated the attempt to modernize the language. But this is not the point for the project. The point is whether this PR brings value or dis-value.  IMHO, it does not bring value for the following reasons:

1. code gets actually harder to read, and not only in a sense of merely reading it, but also in a sense that you may have hard times when trying to understand what is the actual type you are getting the reference to;
2. even just for troubleshooting many times it turns out you have to print the actual type to see what is going on;
3. polymorphic calls are not supported and polymorphism, in general, is not possible, because compiler cannot infer different types with the same code;
4. code actually gets inconsistent, as you'll be having explicit types for fields and "var"s for local variables. This will also confuse newcomers to the code base;
5. explicitly declare type in import is actually positive rather than negative: dependencies are clearer and you know what you are using.

My two cents

PR #288, opened 2023-11-08
rewrite(S3FileSystemProvider): instantiate S3FileAttributes once

guicamest:
*Description of changes:*

The rewrite unifies the `return`s from `readAttributes` taking advantage of the filter option. As a consequence, `asMap()` method from `S3BasicFileAttributes` is not used anymore, so it has been cleaned up.


markjschreiber (approved):
LGTM, thanks!

markjschreiber:
Looks good but the change to using `var` introduced a conflict here that you might need to merge in.

PR #296, opened 2023-11-09
chore: Improve `S3WritableByteChannel` coverage, behaviour, maintenance

guicamest:
*Description of changes:*

This PR includes several unit tests for `S3WritableByteChannel` and `S3TransferUtil` (_new class_), as well as some integration tests that use `Files.write`. In order to ease testing and complexity, the download / upload behaviour from `S3WritableByteChannel` has been delegated to `S3TransferUtil`.

Additionally, **now** when `S3WritableByteChannel` is `close()`d, the temporary file is attempted to be deleted. This should help when writing multiple large files (disk space is freed after each file is copied).


src/test/java/software/amazon/nio/spi/s3/S3SeekableByteChannelTest.java
```diff
@@ -17,12 +17,14 @@
...
-import java.nio.file.StandardOpenOption;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import org.junit.jupiter.api.AfterEach;
+
+import static java.nio.file.StandardOpenOption.CREATE;
+import static java.nio.file.StandardOpenOption.READ;
+import static java.nio.file.StandardOpenOption.WRITE;
```
markjschreiber, line 27: Another todo (for another PR) would be to add a style template so we are consistent about import order, format etc. Currently I am just using IDEAs defaults but a common template would be great.

markjschreiber (approved):

src/main/java/software/amazon/nio/spi/s3/S3WritableByteChannel.java
```diff
@@ -1,13 +1,7 @@
 package software.amazon.nio.spi.s3;
 
-import org.slf4j.Logger;
+import org.checkerframework.checker.nullness.qual.NonNull;
```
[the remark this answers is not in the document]
guicamest: Sure! The annotations are already present (via some transitive dependency). `NonNull` is also used in ReadAheadByteChannel if I recall correctly. Do you mind creating an issue for it? It'd be nice to know whether it is just adding the dependency explicitly or integrate the checks on the build via the checkers plugin (and what checks to include)

src/test/java/software/amazon/nio/spi/s3/S3SeekableByteChannelTest.java
guicamest: Yes, I agree! Do you have any library already in mind?

Issue #300, opened 2023-11-13
Add checkstyle plugin for Gradle


markjschreiber:
https://docs.gradle.org/current/userguide/checkstyle_plugin.html

guicamest:
Any particular style to follow? Some that come to mind:

* [Google Style Guide definition](https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml)
* [Google Style Guide document](https://google.github.io/styleguide/javaguide.html)
* [Checkstyle for AWS SDK V2](https://github.com/aws/aws-sdk-java-v2/blob/master/build-tools/src/main/resources/software/amazon/awssdk/checkstyle.xml)

markjschreiber:
For consistency I guess I should be recommending Checkstyle for AWS SDK V2 :)

PR #311, opened 2023-11-16
build(style): Add Checkstyle plugin for Gradle

guicamest:
*Issue #300*

*Description of changes:*
This PR adds Checkstyle as part of the build process. The configuration is based on [Checkstyle for AWS SDK V2](https://github.com/aws/aws-sdk-java-v2/blob/master/build-tools/src/main/resources/software/amazon/awssdk/checkstyle.xml), without some custom checks on for that specific project.

During `./gradlew build`, the following tasks are executed:
* checkstyleMain
* checkstyleExamples
* checkstyleTest
* checkstyleIntegrationTest

For the test sources only the `header` check is applied.

Given that style is now enforced during build time, many files are formatted / updated 😬 . 

**TODO**
* Add license of checkstyle ⏸️ 

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,6 +14,7 @@ plugins {
     id 'jacoco'
     id 'jvm-test-suite'
     id 'jacoco-report-aggregation'
+    id 'checkstyle'
     id 'com.github.johnrengelman.shadow' version '7.1.2'
     id 'com.palantir.revapi' version '1.7.0'
 }
@@ -164,4 +165,9 @@ jacocoTestReport {
         xml.required = true
         xml.outputLocation = file("build/reports/coverage.xml")
     }
+}
+
+checkstyle {
+    toolVersion = '10.12.5'
+    configDirectory.set(new File(projectDir, ".checkstyle"))
 }
--- a/build.gradle
+++ b/build.gradle
@@ -65,6 +65,7 @@ dependencies {
     implementation 'com.github.ben-manes.caffeine:caffeine:3.1.8'
 
     examplesImplementation project
+    examplesImplementation 'org.slf4j:slf4j-api:2.0.9'
 }
 
 publishing {
--- a/src/examples/java/software/amazon/nio/spi/examples/Main.java
+++ b/src/examples/java/software/amazon/nio/spi/examples/Main.java
@@ -9,12 +9,17 @@ import java.io.IOException;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Demo class that will read text from S3 URIs that you provide on the command line. Your standard AWS credential chain
  * is used for permissions.
  */
 public class Main {
+
+    private static final Logger logger = LoggerFactory.getLogger(Main.class);
+
     /**
      * Demo main method
      *
@@ -24,7 +29,7 @@ public class Main {
     public static void main(String[] args) throws IOException {
 
         if (args.length == 0) {
-            System.err.println("Provide one or more S3 URIs to read from.");
+            logger.error("Provide one or more S3 URIs to read from.");
             System.exit(1);
         }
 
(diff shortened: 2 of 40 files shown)


build.gradle
```diff
@@ -57,13 +58,14 @@ dependencies {
 
     implementation 'software.amazon.awssdk:s3'
     implementation 'software.amazon.awssdk:s3-transfer-manager'
-    implementation 'software.amazon.awssdk.crt:aws-crt:0.28.6'
+    implementation 'software.amazon.awssdk.crt:aws-crt:0.28.0'
```
markjschreiber, line 61: Did you mean to downgrade the version here?

src/examples/java/software/amazon/nio/spi/examples/Main.java
```diff
@@ -9,21 +9,27 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Demo class that will read text from S3 URIs that you provide on the command line. Your standard AWS credential chain
  * is used for permissions.
  */
 public class Main {
+
+    private static final Logger logger = LoggerFactory.getLogger(ListPrefix.class);
```
markjschreiber, line 21: Should use the `Main.class` logger

src/examples/java/software/amazon/nio/spi/examples/Main.java
```diff
@@ -35,10 +41,10 @@ public static void main(String[] args) throws IOException {
             // proves that the correct path type is being used
             assert path.getClass().getName().contains("S3Path");
 
-            System.err.println("*** READING FROM "+path.toUri()+" ***");
+            logger.info("*** READING FROM {} ***", path.toUri());
             Files.readAllLines(path)
-                    .forEach(System.out::println);
-            System.err.println("*** FINISHED READING OBJECT ***\n");
+                .forEach(logger::info);
```
markjschreiber, line 46: Does the Checkstyle doc object to writing to stdout and stderr? Normally that would be good but in the `examples` package we could probably ignore that check? Not a major issue.

src/examples/java/software/amazon/nio/spi/examples/WalkFromRoot.java
```diff
@@ -29,13 +38,13 @@ public static void main(String[] args) throws IOException {
...
         var root = Paths.get(URI.create(bucketName));
-        System.err.println("root.getClass() = " + root.getClass());
+        logger.info("root.getClass() = {}", root.getClass());
 
         var s3 = root.getFileSystem();
 
         for (var rootDir : s3.getRootDirectories()) {
             try (var pathStream = Files.walk(rootDir)) {
-                pathStream.forEach(System.out::println);
+                pathStream.forEach((p) -> logger.info(p.toString()));
```
markjschreiber, line 47: can probably be simplified to `forEach(logger::info)`

markjschreiber (changes requested):
Thanks for adding in a style checker! A couple of comments.

build.gradle
guicamest: No 😅 I'll change it back, thanks for the heads up!

guicamest: Done 👍

src/examples/java/software/amazon/nio/spi/examples/Main.java
guicamest: *Yes*

guicamest: Done 🥳

src/examples/java/software/amazon/nio/spi/examples/Main.java
guicamest: Yes, it is [this part](https://github.com/awslabs/aws-java-nio-spi-for-s3/pull/311/files#diff-9abde9e2da5e93b017c69a8461ca7ba480c567029e141d598ac0fe0dcc2f2b08R323) of our config (copied from aws-sdk).

I thought it would be nice to have it in the examples also, so that it comes handy to users that already use logger

src/examples/java/software/amazon/nio/spi/examples/WalkFromRoot.java
guicamest: No 😞 `p` is a `Path`, not a `String`. And `logger.info` takes a `String`

PR #342, opened 2023-12-25
chore(S3BasicFileAttributes): Do IO in methods that throw `IOException`

guicamest:
*Description of changes:*

The implementation of S3BasicFileAttributes does IO whenever `size()`, `lastModified()` or `fileKey()` are invoked. When there are network problems, `RuntimeException`s are thrown, since the `BasicFileAttributes` *nio* interface does not declare any expected exceptions to be thrown when calling these methods.

In order to conform better with the contract specified by the *nio* interfaces, the IO done in `size()`, `lastModified()` and `fileKey()` methods is now done before constructing the attributes: when `readAttributes` from the *provider* or the *attributeview* is called. `readAttributes` in those interfaces specify that it can throw `IOException` and therefore seems a better place where to do the IO to get the object metadata.

As a side effect, requests to S3 are **reduced** whenever a user of a library needs **more than 1 attribute** of a file; for example `size` and `lastModified`.

## References:
* [BasicFileAttributes javadoc](https://docs.oracle.com/javase/8/docs/api/java/nio/file/attribute/BasicFileAttributes.html)
* [BasicFileAttributeView.readAttributes javadoc](https://docs.oracle.com/javase/8/docs/api/java/nio/file/attribute/BasicFileAttributeView.html#readAttributes--)
* [FileSystemProvider.readAttributes javadoc](https://docs.oracle.com/javase/8/docs/api/java/nio/file/spi/FileSystemProvider.html#readAttributes-java.nio.file.Path-java.lang.Class-java.nio.file.LinkOption...-)

### DIFF
--- a/src/main/java/software/amazon/nio/spi/s3/S3BasicFileAttributeView.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3BasicFileAttributeView.java
@@ -5,6 +5,7 @@
 
 package software.amazon.nio.spi.s3;
 
+import java.io.IOException;
 import java.nio.file.attribute.BasicFileAttributeView;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.nio.file.attribute.FileTime;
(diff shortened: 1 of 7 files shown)


markjschreiber (approved):
Nice! Thanks

Issue #353, opened 2024-01-09
AWS_REGION is ignored => region is locked to us-east-1


bbilger:
It's currently not possible to change the region e.g. via AWS_REGION or whatever the provider chain would consider.
The problem is that `Region.US_EAST` is set on the builder when none is provided; and none can be provided currently in anyway. Thus it is locked to that region.
https://github.com/awslabs/aws-java-nio-spi-for-s3/blob/935a548f8899b35b25b3377c761d536a34477156/src/main/java/software/amazon/nio/spi/s3/S3ClientProvider.java#L300

This problem happens primarily for 3rd party providers but it can also happen for AWS buckets when the region cannot be determined.
https://github.com/awslabs/aws-java-nio-spi-for-s3/blob/935a548f8899b35b25b3377c761d536a34477156/src/main/java/software/amazon/nio/spi/s3/S3ClientProvider.java#L170

markjschreiber:
Can you check if this is resolved by PR #400 ?

markjschreiber:
Closing as PR 400 removes the defaulting to region `us-east-1`.

bbilger:
> Can you check if this is resolved by PR #400 ?

Looks good, thanks - sorry for the delay.

PR #354, opened 2024-01-09
[#353] leave region empty if none is provided

bbilger:
...and as such leave the decision to the provider chain (e.g. env var AWS_REGION)

*Issue #, if available:*

cannot change region in any way

*Description of changes:*

if no region is provided then the decision will be made by the provider chain

### DIFF
--- a/src/main/java/software/amazon/nio/spi/s3/S3ClientProvider.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3ClientProvider.java
@@ -253,11 +253,14 @@ public class S3ClientProvider {
         S3AsyncClientBuilder builder
     ) {
         var region = getRegionFromRegionName(regionName);
-        logger.debug("bucket region is: '{}'", region.id());
+        // no region provided => leave it to the provider chain to figure it out (e.g. env var AWS_REGION)
+        if (region != null) {
+            logger.debug("bucket region is: '{}'", region.id());
+            asyncClientBuilder.region(region);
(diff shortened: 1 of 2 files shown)


stefanofornari:
It looks good to me, but I may miss the background that lead to pick a default region. @markjschreiber ?

markjschreiber:
> It looks good to me, but I may miss the background that lead to pick a default region. @markjschreiber ?

I believe S3 predates the concept of AWS regions which is why you get lots of quirks like bucket names are global but buckets have regions. This means that some SDK APIs that tell you the region of the bucket need to be directed to the `us-east-1` endpoint. The sigv4 signature protocol requires a signature from a client configured for the same region that you are making a request to. So, if the bucket `my-bucket` lives in `us-west-2` but you don't know that you won't be able to make requests to it until you figure out which region your client should be set to. Therefore we need a "region discovery client" that is always `us-east-1` to do the region lookup. Once you find out that the region is `us-west-2` then you need make a client with that region to make further requests to that bucket.

So, having `us-east-1` as a default for `null` is a bug (except for region lookups) and `null` should fall back to the provider chain.

Another weird quirk is that the region lookup for a bucket in the `us-east-1` region will return an empty string. This is possibly a legacy from when there was no regions and everything was `us-east-1`? The S3 API can probably not change this now as it would be a breaking change for legacy code.

markjschreiber:
Also resolved by PR #400

Issue #293, opened 2023-11-09
S3Path: relativize can return an absolute path


thsandu:
We are relying in a part of our implementation on the behavior of the Path.relativize() to return a relative path. I discovered that the S3Path gives an absolute path back, if the relativize is executed on the root path. I checked again against windows and jimfs implementation:

```
@Test
public void relativizeAbsolutePathOnRoot() {
    Path path = Paths.get("C:\\");
    Path path2 = Paths.get("C:\\foo\\bar\\");
    assertFalse(path.relativize(path2).isAbsolute());
    
    FileSystem linux = Jimfs.newFileSystem("testme", Configuration.unix());
    Path pathJim = linux.getPath("/");
    Path pathJim2 = linux.getPath("/foo/bar/");
    assertFalse(pathJim.relativize(pathJim2).isAbsolute());
}
```
Here again a matching unit test: https://github.com/thsandu/aws-java-nio-spi-for-s3-issues-with-s3x/blob/2cbb6d27b8adde8499f49a146eedce870dec1ec6/src/test/java/software/amazon/nio/spi/s3/S3PathTest.java#L293

markjschreiber:
So does
```
    Path pathJim = linux.getPath("/");
    Path pathJim2 = linux.getPath("/foo/bar/");
```

result in a path equivalent to "./foo/bar/"?

markjschreiber:
Can you check if PR #394 has fixed this?

markjschreiber:
This test seems to pass with PR #394.

```
    @Test
    void testRelativizeAgainstRoot() {
        var s3Path = S3Path.getPath(fileSystem, "/");
        var relativePath = s3Path.relativize(S3Path.getPath(fileSystem, "/a/b/c/d"));
        assertEquals("a/b/c/d", relativePath.toString());
        assertFalse(relativePath.isAbsolute());

        var emptyS3Path = S3Path.getPath(fileSystem, "");
        relativePath = emptyS3Path.relativize(S3Path.getPath(fileSystem, "a/b/c/d"));
        assertEquals("a/b/c/d", relativePath.toString());
        assertFalse(relativePath.isAbsolute());
    }
```

markjschreiber:
I have added some additional tests to prevent a regression. Feel free to open a new issue if this is still a problem.

PR #394, opened 2024-03-03
fix for #293

stefanofornari:
*Issue #, if available:*
#293 

*Description of changes:*
The current implementation violates the intended behaviour or relativize when relativizing a path to the root. The correct behaviour should be like:

```
Path root = FileSystems.getDefault().getRootDirectories().iterator().next();
Path path = FileSystems.getDefault().getPath("/a/b");
Path relative = root.relativize(path); // -> a/b, isAbsolute() == false

```

Side note: the use of var is particularly disappointing in `S3Path` where it should be clear if an object is a `Path` or a `S3Path`

### DIFF
--- a/src/main/java/software/amazon/nio/spi/s3/S3Path.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3Path.java
@@ -257,10 +257,6 @@ class S3Path implements Path {
         }
 
         var path = String.join(PATH_SEPARATOR, pathRepresentation.elements().subList(beginIndex, endIndex));
-        if (this.isAbsolute() && beginIndex == 0) {
-            path = PATH_SEPARATOR + path;
-        }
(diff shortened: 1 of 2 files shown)


src/main/java/software/amazon/nio/spi/s3/S3Path.java
```diff
@@ -257,10 +257,6 @@ public S3Path subpath(int beginIndex, int endIndex) {
         }
 
         var path = String.join(PATH_SEPARATOR, pathRepresentation.elements().subList(beginIndex, endIndex));
```
markjschreiber, line 259: `path` can probably be `final`

markjschreiber (approved):
Thanks for this fix! I think `path` can probably be made `final` now but not critical.

Issue #399, opened 2024-03-05
S3Exception: the region 'eu-west-1' is wrong; expecting 'us-east-1'


heuermh:
Note may be a duplicate of #353.

With version 1.2.4, a warning is logged when the bucket region does not match the account region, but the downloaded succeeds
```
$ foo -i s3://ngi-igenomes/igenomes/Homo_sapiens/Ensembl/GRCh37/Annotation/Genes/genes.bed
[WARN] S3ClientStoreProvider - Unable to determine the region of bucket: 'ngi-igenomes'. Generating
a client for the profile region.
```

With version 2.0.0-rc1, `S3Exception` is thrown
```
$ foo -i s3://ngi-igenomes/igenomes/Homo_sapiens/Ensembl/GRCh37/Annotation/Genes/genes.bed
java.lang.RuntimeException: java.util.concurrent.ExecutionException:
software.amazon.awssdk.services.s3.model.S3Exception: The authorization header is malformed; the region
'eu-west-1' is wrong; expecting 'us-east-1' (Service: S3, Status Code: 400, Request ID: ..., Extended Request ID: .../...=)
	at software.amazon.nio.spi.s3.S3ClientProvider.generateClient(S3ClientProvider.java:133)
	at software.amazon.nio.spi.s3.S3FileSystem.readClient(S3FileSystem.java:422)
	at software.amazon.nio.spi.s3.S3SeekableByteChannel.<init>(S3SeekableByteChannel.java:68)
	at software.amazon.nio.spi.s3.S3SeekableByteChannel.<init>(S3SeekableByteChannel.java:40)
	at software.amazon.nio.spi.s3.S3FileSystemProvider.newByteChannel(S3FileSystemProvider.java:226)
	at java.base/java.nio.file.Files.newByteChannel(Files.java:379)
	at java.base/java.nio.file.Files.newByteChannel(Files.java:431)
	at java.base/java.nio.file.spi.FileSystemProvider.newInputStream(FileSystemProvider.java:420)
	at java.base/java.nio.file.Files.newInputStream(Files.java:159)
	...
Caused by: java.util.concurrent.ExecutionException: software.amazon.awssdk.services.s3.model.S3Exception: The authorization header is malformed; the region 'eu-west-1' is wrong; expecting 'us-east-1' (Service: S3, Status Code: 400, Request ID: ..., Extended Request ID: .../...=)
[...]

PR #400, opened 2024-03-06
fix for issue 399

markjschreiber:
*Issue #, if available:*
#399 
*Description of changes:*
removes unnecessary creation of default client when a client already exists


markjschreiber:
Should also fix #353

PR #413, opened 2024-03-17
chore: Add logback for tests

guicamest:
*Description of changes:*
When running tests a developer cannot see any logs from the library. This PR adds `logback-classic` as a dependency to the test sourcesets (test / integrationTest) and shouldn't contribute to the final size of the library


markjschreiber (approved):
Looks good. Thanks!

PR #415, opened 2024-03-17
chore: Cleanups on `S3ClientProvider`

guicamest:
*Description of changes:*

The `readClient` method is unused and removed in the PR. This allows to cleanup a lot of code related to whether a client to be generated should be crt (always) or not (never).


src/main/java/software/amazon/nio/spi/s3/S3ClientProvider.java
```diff
@@ -117,64 +71,48 @@ S3AsyncClient universalClient() {
         return DEFAULT_CLIENT;
     }
 
-
     /**
      * Generates a sync client for the named bucket using the provided location
```
markjschreiber, line 75: Generates as Async Client

markjschreiber (approved):
Great to have this cleaned up ahead of 2.0

Issue #458, opened 2024-05-15
Improve efficiency of `Files.walk()`


markjschreiber:
Walking a tree using `Files.walk()` is slow in the current implementation and likely uses too many calls to S3. The backend implementation may need to be more eager in collecting paths in groups.

markjschreiber:
Investigations reveal that calling `Files.walk()` will require many calls to S3 including a Head for every object (to get file metadata). These calls coming from multiple places is resulting in a large number of requests to generate new S3 clients. 

A potentially simple way to speed up client creation would be to retain a cache of known regions for clients to prevent having to make a HeadBucket request per client creation.

We could also re-examine the idea of a client pool or cache with appropriate cautions about auto-closed clients.

PR #466, opened 2024-05-20
fix: improve speed of `Files.walk()`

markjschreiber:
*Issue #, if available:*
#458 

*Description of changes:*
caches results of expensive calls to determine the bucket region and caches clients for buckets to reduce the large expense of creating a client. Clients are meant to be re-used according to SDK docs and are thread safe. Because clients can be "auto-closed", Cached clients are wrapped so that their "closed" status can be recorded and queried. Generally clients should not be closed unless you know you will no longer need them (due to the cost of creating them). 


markjschreiber:
fixes #458

PR #469, opened 2024-05-27
Add `FORCE_PATH_STYLE` into the defaults of S3NioSpiConfiguration in order to allow it to be modified via environment variables

oeph:
Due to the "auto-create" of the file systems, we experience cases, where we cannot set the FORCE_PATH_STYLE method to `true`.

*Description of changes:*
By adding the `s3.spi.force-path-style` option to the defaults, it will get picked up by the default handling and can be modified via environment variables. The default is set to `false`.


markjschreiber (changes requested):
Looks good however you will need to make some changes in the unit tests that are failing.

PR #480, opened 2024-06-20
Use `ConcurrentHashMap` to protect `FS_CACHE`

### LINKED ISSUE #474: `java.util.ConcurrentModificationException` after upgrading to 2.0.1 from 1.2.4
Hi everyone,

when bumping to 2.0.1, we started seeing the following:

```
    java.util.ConcurrentModificationException
        at java.base/java.util.HashMap.computeIfAbsent(HashMap.java:1229)
        at software.amazon.nio.spi.s3.S3FileSystemProvider.getFileSystem(S3FileSystemProvider.java:775)
        at software.amazon.nio.spi.s3.S3FileSystemProvider.getPath(S3FileSystemProvider.java:251)
        at java.base/java.nio.file.Path.of(Path.java:209)
        [...] ommited rest of stacktrace
```

The exception is raised from the following (abbreviated) code:

```
listOfFiles
  .stream()
  .map(file -> Path.of(URI.create(file.path)).getFileName().toString())
  .toList();
```
which is executed in three [cucumber-jvm](https://github.com/cucumber/cucumber-jvm) 7.15.0 test scenarios which are executed in parallel. Configured by:

```
cucumber.execution.parallel.enabled = true
```

in junit-platform.properties.

I haven't seen anything in the 2.x changelogs which caught my eye. Any help is appreciated, thx!

hanneskaeufler:
Closes #474

*Issue #, if available: #474

*Description of changes:*


markjschreiber:
The checks are passing now and the code looks good. Do you want to remove the draft status?

hanneskaeufler:
> The checks are passing now and the code looks good. Do you want to remove the draft status?

Sure! To be transparent though: I have not verified whatsoever that this solves the problem I was seeing in the referenced issue. It was more just a hunch that crossed my mind and I wanted to get to paper. I see no particular _risk_ with this either, so we might as well get it in? Not sure about test automation here ... :/

markjschreiber:
> > The checks are passing now and the code looks good. Do you want to remove the draft status?
> 
> Sure! To be transparent though: I have not verified whatsoever that this solves the problem I was seeing in the referenced issue. It was more just a hunch that crossed my mind and I wanted to get to paper. I see no particular _risk_ with this either, so we might as well get it in? Not sure about test automation here ... :/

Agreed, there is no risk. There might be a slight performance drop but should be trivial compared to the network IO we have to do with S3. I don't know that you could unit test this other than to ensure that the Map is an instance of ConcurrentMap just to make sure nobody changes it back again without good reason.

I assume you would want to test it with your parallel test framework and it is always a good idea to try running the apps in the `examples` dir to ensure nothing weird happens.

riklopfer:
I just encountered the same issue in a similar setting to @hanneskaeufler Is there a plan to merge this fix in?

markjschreiber:
> I just encountered the same issue in a similar setting to @hanneskaeufler Is there a plan to merge this fix in?

I think it makes sense. The Branch has going a bit stale. I will try rebasing onto main.

markjschreiber (approved):
Thanks @hanneskaeufler

PR #485, opened 2024-06-25
fix: Let getFileAttributeView() return null for unsupported view class.

chbriem:
*Issue #, if available:*
No issue created this time. If needed, I can create one later.

*Description of changes:*
Observed when trying to use JDK's `Files.move()` method. This method uses the `java.nio.file.CopyMoveHelper#copyToForeignTarget` method. The implementation tries to get the `PosixFileAttributeView` (in line 109). This fails for the S3 implementation with an `UnsupportedOperationException`, which is not caught by JDK's code, resulting in the move operation to fail.

Having read the method description, I'm of the opinion that the method should return `null` in case of an unsupported view type. This would at least be compliant with the method comment in the `copyToForeignTarget` method (line 108):

> // retrieve source posix view, null if unsupported

I've adapted the test accordingly.


markjschreiber (approved):
Looks good. Thanks!

PR #498, opened 2024-07-17
fix: fixes refreshing expired credentials by preventing DefaultCredentialsProvider from being autoclosed

### LINKED ISSUE #494: java.lang.IllegalStateException: Connection pool shut down
I have run into a quite unusual issue where after the service has been running for 1 hour an error is thrown from [here](https://github.com/awslabs/aws-java-nio-spi-for-s3/blob/eb9039829bef5f015a5726f0f137a9e01879409a/src/main/java/software/amazon/nio/spi/s3/S3BasicFileAttributes.java#L241) with a completely useless error message: 
```
software.amazon.awssdk.core.exception.SdkClientException: Failed to send the request: A callback has reported failure.
```

This only occured when using temporary credentials, both with `AssumeRoleWithWebIdentity` and `AssumeRole` requests. I had a bit of a dig around and eventually got a debugger into the right place and found the following error message thrown when trying to refresh the credentials `java.lang.IllegalStateException: Connection pool shut down`. From [this](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/troubleshooting.html#faq-connection-pool-shutdown-exception) troubleshooting guide it suggests the error is thrown when there is an attempt to use a closed `DefaultCredentialsProvider`.

The only place I could find that was closing the DefaultCredentialsProvider was [here](https://github.com/awslabs/aws-java-nio-spi-for-s3/blob/eb9039829bef5f015a5726f0f137a9e01879409a/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java#L387). I was able to fudge in a fix by pulling the `client` allocation out of the `try-with-resource` block and this seems to work. I'm not sure if this is the correct fix as there is likely some interaction here that I don't quite understnad.
```
diff --git a/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java b/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
index 55254d2..29ed0f4 100644
--- a/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
@@ -384,7 +384,8 @@ public class S3FileSystemProvider extends FileSystemProvider {
         var timeOut = TIMEOUT_TIME_LENGTH_1;
         final var unit = MINUTES;

-        try (S3AsyncClient client = s3Directory.getFileSystem().client()) {
+        var client = s3Directory.getFileSystem().client();
+        try {
             client.putObject(
                 PutObjectRequest.builder()
                     .bucket(s3Directory.bucketName())
```

Sorry I don't have full stacktraces, reproducing this takes an age and I've been chasing this down for so long my brain is starting to melt! The easiest way to reproduce the issue is to set up an AWS profile that assumes a role, and then wait for an hour.

sberss:
Fixes #494

Pulls the client instantiation out of the `try-with-resource` block to stop it being closed and therefore not usable when refreshing expired credentials.

### DIFF
--- a/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
@@ -384,8 +384,8 @@ public class S3FileSystemProvider extends FileSystemProvider {
         var timeOut = TIMEOUT_TIME_LENGTH_1;
         final var unit = MINUTES;
 
-        try (S3AsyncClient client = s3Directory.getFileSystem().client()) {
-            client.putObject(
+        try {
+            s3Directory.getFileSystem().client().putObject(
                 PutObjectRequest.builder()
                     .bucket(s3Directory.bucketName())
                     .key(directoryKey)


markjschreiber (approved):
Thanks for this!

sberss:
No worries! Any chance I could get this and #499 in a new release?

PR #499, opened 2024-07-17
Correctly handle copy and delete operations on single objects

### LINKED ISSUE #497: Copy and delete functionality broken for objects sharing prefix
The [copy](https://github.com/awslabs/aws-java-nio-spi-for-s3/blob/01542979007859a51adfcc0aef9338e8a4fa4ade/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java#L459) method seems to be behaving incorrectly when performing a bucket to bucket copy for a single object. Say you have the following files in an S3 bucket:

```
s3://bucket-A/foo
s3://bucket-A/foo.txt
```

If you were to do a `fs.copy("s3://bucket-A/foo", "s3://bucket-B/foo")`, both `foo` and `foo.txt` are copied to `bucket-B`, which I believe is incorrect behaviour. This is becuase the function uses the source key as a prefix to ListBuckets [here](https://github.com/awslabs/aws-java-nio-spi-for-s3/blob/01542979007859a51adfcc0aef9338e8a4fa4ade/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java#L862) without checking whether the source is a directory.

I believe this is also the case for delete operations.

sberss:
Fixes #497

The copy and delete code was currently treating `/foo/` and `/foo` as a directory, which in the S3 context is incorrect (as perfectly summed up [here](https://github.com/awslabs/aws-java-nio-spi-for-s3/blob/01542979007859a51adfcc0aef9338e8a4fa4ade/src/main/java/software/amazon/nio/spi/s3/PosixLikePathRepresentation.java#L146)). If the given path is not a directory, it should not be listing objects from S3 and returning all those that match the prefix. Rather, it should just be performing the copy/delete operation on that given object. This change adds a check for whether the given source is a directory and acts accordingly.

### DIFF
--- a/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
@@ -419,7 +419,9 @@ public class S3FileSystemProvider extends FileSystemProvider {
         var timeOut = TIMEOUT_TIME_LENGTH_1;
         final var unit = MINUTES;
         try {
-            var keys = getContainedObjectBatches(s3Client, bucketName, prefix, timeOut, unit);
+            var keys = s3Path.isDirectory() ?
+                    getContainedObjectBatches(s3Client, bucketName, prefix, timeOut, unit)
+                    : List.of(List.of(ObjectIdentifier.builder().key(prefix).build()));
 
             for (var keyList : keys) {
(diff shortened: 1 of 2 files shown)


markjschreiber:
LGTM!

markjschreiber (approved):

markjschreiber:
New release is out (2.0.4)

sberss:
Thank you!

Issue #597, opened 2025-01-16
Configuration scope problem with fix for issue #513.


sielenk-yara:
The fix for issue #513 (pr #517) adds new configuration values for timeouts.
These values are solely used in the S3FileSystemProvider.
But they are held by the S3FileSystem and are propagated to the provider on creation.
Hence the same provider will "inherit" the configured values from the latest filesystem created.

As long as they are the same, that's not a problem.
But when working with several buckets this might actually change the behavior of existing paths on the fly.

sielenk-yara:
Alternatively there could be a separate, bucket-independent configuration for the provider.
Or at least an information flow down from provider to filesystem, not the other way around as it was implemented: default config in the provider, to be passed down and potentially modified in an overwritten factory method, but not propagated back to the provider.

PR #598, opened 2025-01-16
fix: removed the configuration from the filesystem provider and used the one in the filesystem instead.

sielenk-yara:
This is a fix for issue #597.

### DIFF
--- a/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
@@ -98,17 +98,6 @@ public class S3FileSystemProvider extends FileSystemProvider {
     static final String SCHEME = "s3";
     private static final Map<String, S3FileSystem> FS_CACHE = new ConcurrentHashMap<>();
 
-    /**
-     * This variable holds the configuration for the S3 NIO Service Provider Interface (SPI).
-     * It is used to manage and handle the configuration details required for interaction
-     * with S3 NIO services.
-     *
-     * @deprecated This variable is deprecated and may be removed in future versions.
-     *             Consider using updated configuration mechanisms if available.
-     */
-    @Deprecated
-    protected S3NioSpiConfiguration configuration = new S3NioSpiConfiguration();
-
     private final Logger logger = LoggerFactory.getLogger(this.getClass().getName());
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
```diff
@@ -801,15 +804,6 @@ public void setAttribute(Path path, String attribute, Object value, LinkOption..
         throw new UnsupportedOperationException("s3 file attributes cannot be modified by this class");
     }
 
-    /**
-     * Set custom configuration. This configuration is referred to for API timeouts
-     *
-     * @param configuration    The new configuration containing the timeout info
-     */
-    public void setConfiguration(S3NioSpiConfiguration configuration) {
```
markjschreiber, line 809: removing a public method is a breaking change to the API. If possible it should be reimplemented or alternatively deprecate the method and update the docs with a recommendation on a better alternative.

markjschreiber (changes requested):
Looks good. Some things to change:

1. The public method should be deprecated rather than removed (a breaking change).
2. Add unit tests to demonstrate that the configuration has been set to prevent regressions in the future.

src/main/java/software/amazon/nio/spi/s3/S3FileSystemProvider.java
sielenk-yara: Yes, that is a breaking change.
But given the rest of the change, I see no reasonable way to keep that method. The field it used to set is gone, and there is no getter for it. The code that used the field gets the configuration from the `S3FileSystem`s now and doesn't use it either.

I could put in a deprecated dummy method that just accepts the configuration but that would break client code in more subtle ways with just a warning on compilation.

Alternatively I could keep the field, make it null by default and use it only, if it has been set. That would still change the semantics (no more updates from the filesystems), but probably keep code using it happy(ish).

What to do?

markjschreiber: I think the goal should be to deprecate it and if we can't make it any better, at least make it no worse. Document in the method comments why it shouldn't be used and what would be a better alternative.

Next major version we can make breaking changes by removing deprecated methods. With minor versions we try to prevent breaking changes.

sielenk-yara: Okay, then I'll start a new PR for that an leave this one for later.

markjschreiber:
We could definitely deprecate the old method. I'm also wondering if there is a way that the old method can call the new method in the `FileSystemProvider`?

Either way, even without a major version change we should deprecate the old method and tell people to use the new one instead and then remove the deprecated method with the next major version. Especially as the old method has limited utility.

Issue #599, opened 2025-01-17
There is a "region" value in S3NioSpiConfiguration, but it is not propagated to the S3 client.


sielenk-yara:
In tests I am writing I have no global AWS region or credentials configured.
Instead I set them in the configuration map of the `S3FileSystem` but noticed that while the credentials are used, the region is not.

PR #600, opened 2025-01-17
fix: propagate the configured region to the S3 client

sielenk-yara:
This is a fix for issue #599.


markjschreiber (changes requested):
Looks good, however I think a unit test should be added to assert that the region is set.

sielenk-yara:
Sorry, didn't notice I need to re-request a review. This one should be good to go.

markjschreiber (approved):
Looks great. Thanks for the contribution

Issue #601, opened 2025-01-17
Doing anything with the S3NioSpiConfiguration is difficult and error prone.


sielenk-yara:
I am trying to pass credential and region information to the S3 client in my code. Since this happens in test code, I can't use global locations like the SystemProperties without incurring severe race condition risks.

But due to #227, this is almost impossible. I ended up with this Kotlin code:

```Kotlin
            val provider: S3FileSystemProvider =
                object : S3FileSystemProvider() {
                    override fun getFileSystem(uri: URI): FileSystem {
                        val fileSystem = super.getFileSystem(uri)

                        // since super.getFileSystem puts the new filesystems configuration
                        // into this protected field, it can be accessed like this:
                        configuration.withRegion("eu-central-1")
                        configuration.withCredentials(credentials)

                        return fileSystem
                    }
                }

```

This only works because `super.getFileSystem(uri)` updates a protected field in the Provider.
The configuration is actually owned by the fileSystem object, so this is problematic (see #597).

I think #227 went way too far in hiding things. I am inside the `S3FileSystemProvider`, and I *know* the function will return a `S3FileSystem` which could (should?) allow access to its configuration. But the configuration getter is now package private, so not accessible from my code.

I do agree that the actual implementation classes should be hidden. But that can be done by adding abstract classes that expose only what is necessary without revealing too much. Just falling back to the `java.nio.file.spi` classes hides too much. Interfaces would be nice, but the JRE already has abstract classes, so that is not possible.

markjschreiber:
Making the configuration more visible probably makes sense, however I am wondering about your use case.

Normally in testing you would mock the S3 bucker (FileSystem) or for integration testing you might want to use a containerized S3 implementation like we do with this projects integration test suite. If you perform integration testing with a real S3 bucket then you will want to ensure that any credentials you provide are temporary credentials. Best practice should be to never use static credentials in any situation.

Also, if you are testing from an EC2 or ECS/EKS container then you shouldn't need to provide credentials or region and these will be obtained from the machines/ containers profile and have access according to the IAM policy attached to them.

Perhaps I am not understanding the situation fully but generally you can avoid the need to use in code configuration.

sielenk-yara:
I know, and I agree...
But I am working on a legacy application and try to get that connected to the cloud while staying within the framework provided/required by corporate IT. So please don't judge me to harshly.
[...]

PR #602, opened 2025-01-17
fix: S3FileSystem configuration visibility and name update

sielenk-yara:
This is a partial fix for issue #601.

### DIFF
--- a/src/main/java/software/amazon/nio/spi/s3/S3FileSystem.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3FileSystem.java
@@ -392,6 +392,22 @@ public class S3FileSystem extends FileSystem {
     }
 
     /**
+     * Returns the configuration object passed in the constructor or created
+     * by default.
+     *
+     * @return the configuration object for this file system
+     *
+     * @deprecated This method is deprecated and may be removed in a future release.
+     *   Use {@link #getConfiguration()} instead to retrieve the configuration object,
+     *   as it provides the same functionality with better naming conventions that
+     *   align with standard practices.
+     */
+    @Deprecated
+    public S3NioSpiConfiguration configuration() {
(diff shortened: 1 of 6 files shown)


src/main/java/software/amazon/nio/spi/s3/S3FileSystem.java
```diff
@@ -387,7 +387,7 @@ public WatchService newWatchService() {
      *
      * @return the configuration object for this file system
      */
-    S3NioSpiConfiguration configuration() {
```
markjschreiber, line 390: I wouldn't rename the method, just make it public. It's a potentially breaking change.

markjschreiber (changes requested):
If you leave the method name the same and just make it public you avoid possible breaking changes and only need to modify 1 file.

Is there a need to follow a bean naming convention?

src/main/java/software/amazon/nio/spi/s3/S3FileSystem.java
sielenk-yara: Hm. It used to be public in an earlier version, so I do see your point.
But the breaking change already happened in making it package private, didn't it?
My reason to change it wasn't the bean naming convention, at least not directly. It was Kotlin, which maps bean convention getter/setter pairs to properties.

I'll change it, though.

markjschreiber:
Hi @sielenk-yara, let me know what the current status is with these open PR's. If there is anything you'd like me to review or merge and in any particular order, let me know.

src/main/java/software/amazon/nio/spi/s3/S3FileSystem.java
sielenk-yara: Subversively kept my changes and re-added the backward method as deprecated.
Will that do?

PR #612, opened 2025-02-05
fix: deprecated the configuration field and setter

sielenk-yara:
This is an intermediate 'fix' for issue #597 until API breaking changes are possible.


markjschreiber:
Why not add the new mechanism to set configuration at the same time as deprecating the old? Adding new methods is backward compatible and will not break anyones legacy code.

markjschreiber (approved):
Thanks for this. Just thinking, we could also open a new branch for a tentative V3 with breaking changes if required.

PR #625, opened 2025-02-28
fix: made S3FileSystemProvider#newFileSystem better match its documentation

sielenk-yara:
The documentation states "May be null or empty." but that was clearly not the case.

Added a test for the fix which expects call to not fail with anything but an `IOException`.


markjschreiber (approved):
Thanks for making this consistent with the docs!

PR #634, opened 2025-03-04
Update JUnit Jupiter to 5.11.4

arouel:
To match the JUnit Jupiter version used by `mockito-junit-jupiter` we should set the JUnit Jupiter of this project to `5.11.4`. Additionally, the logging implementation `logback-classic` is set to `runtimeOnly` scope (for tests), because no one should use the code directly while writing tests.


markjschreiber (approved):
Looks Good!

PR #635, opened 2025-03-04
Transport system properties to S3 NIO configuration

arouel:
According to documentation the endpoint protocol can be configured via system properties, which is not working at the moment. This change makes this possible.


arouel:
@markjschreiber what do you think about this change?

markjschreiber:
Thanks for catching that. Can you also add a unit test so we don't regress here?

arouel:
@markjschreiber I double-checked the implementation with tests and it seems to work as expected, therefore I close the PR.

PR #636, opened 2025-03-06
Add full object checksum when uploading a locally cached file to S3

arouel:
I want to add a full object checksum based on the `CRC32C` or `CRC64NVME` algorithm before uploading the locally cached file to S3.


markjschreiber (approved):
Nice addition, thanks!

PR #639, opened 2025-03-11
Base64 encode the checksum

arouel:
We've figured, that the integrity checksum must be Base64 encoded, but the existing integration tests didn't show the problem due to a wrong setup. Additionally, we need an updated LocalStack docker image, to be able to verify the `CRC64NVME` checksum algorithm.

While testing, we also observed debug messages with stack traces that were caused by head-object requests, where it is expected that an `NoSuchKeyException` occurs. Therefore, we decided to handle this case explicitly and avoid logging (see `S3FileSystemProvider.java` line 901).


markjschreiber (approved):
Thanks for the update

PR #640, opened 2025-03-12
Avoid unnecessary head-object request

arouel:
We want to avoid an additional head-request to S3 when we know the file exists. This reduces the latency and cost when opening a `SeekableByteChannel` or `FileChannel` with the open option `WRITE`.

### DIFF
--- a/src/main/java/software/amazon/nio/spi/s3/S3WritableByteChannel.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3WritableByteChannel.java
@@ -18,10 +18,12 @@ import java.nio.file.StandardOpenOption;
 import java.util.HashSet;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeoutException;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import software.amazon.awssdk.services.s3.S3AsyncClient;
+import software.amazon.awssdk.services.s3.model.S3Exception;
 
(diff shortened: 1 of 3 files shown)


src/main/java/software/amazon/nio/spi/s3/S3WritableByteChannel.java
```diff
@@ -44,18 +46,31 @@ class S3WritableByteChannel implements SeekableByteChannel {
...
             tempFile = Files.createTempFile("aws-s3-nio-", ".tmp");
-            if (exists) {
-                s3TransferUtil.downloadToLocalFile(path, tempFile);
+            // this complicated download handling is the result of
+            // avoiding an existence check with a head-object request
+            if (!options.contains(StandardOpenOption.CREATE_NEW)) {
+                try {
+                    s3TransferUtil.downloadToLocalFile(path, tempFile);
+                } catch (CompletionException e) {
+                    var cause = e.getCause();
```
markjschreiber, line 61: You could switch on the cause being an instance of NoSuchKeyException rather than using the status code. This would make the logic more clear.

markjschreiber (changes requested):
Return code 404 equates to a NoSuchKeyException. Using that in your catch logic should make the handling simpler.

src/main/java/software/amazon/nio/spi/s3/S3WritableByteChannel.java
markjschreiber: Got it. Local stack might not be totally faithful to the API in this case. I will resolve the comment and merge. Thanks for the PR!

markjschreiber (approved):

PR #641, opened 2025-03-12
Use a temporary directory per file system and preserve the directory structure locally

arouel:
To be able to better understand which file I'm messing around with while debugging locally, I find it really helpful when the directory structure is preserved as it exists on S3. Therefore, I suggest to have a temporary directory per `S3FileSystem` instance and locally cached S3 files are mirrored there with their key in correspondence to the bucket location.

### DIFF
--- a/src/main/java/software/amazon/nio/spi/s3/S3FileSystem.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3FileSystem.java
@@ -9,6 +9,7 @@ import static software.amazon.nio.spi.s3.Constants.PATH_SEPARATOR;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.nio.channels.Channel;
 import java.nio.file.ClosedFileSystemException;
 import java.nio.file.DirectoryStream;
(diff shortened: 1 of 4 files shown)


src/main/java/software/amazon/nio/spi/s3/S3FileSystem.java
```diff
@@ -56,14 +56,16 @@ public class S3FileSystem extends FileSystem {
...
+    private final Path temporaryDirectory;
 
     /**
      * Create a filesystem that represents the bucket specified by the URI
      *
      * @param provider the provider to be used with this fileSystem
      * @param config   the configuration to use; can be null to use a default configuration
+     * @param temporaryDirectory the local temporary directory for the S3 object
      */
-    S3FileSystem(S3FileSystemProvider provider, S3NioSpiConfiguration config) {
```
markjschreiber, line 66: To maintain backward compatibility can we keep the old constructor and add this new one as well?

markjschreiber (changes requested):

src/main/java/software/amazon/nio/spi/s3/S3FileSystem.java
arouel: Yes, I changed it.

markjschreiber (approved):

markjschreiber:
Looks good! Thanks

PR #643, opened 2025-03-17
Provide S3 specific `OpenOption` to add a HTTP `Range` header

arouel:
I want to partially get an S3 object when opening a `SeekableByteChannel` or `FileChannel`. When a HTTP `Range` header is specified the HEAD object request is omitted.


markjschreiber (approved):
Thanks!

PR #644, opened 2025-03-17
Improve exception handling on download/upload

arouel:
An exception while downloading from S3 or uploading to S3 is not traceable to the caller, because it is created in another thread. To make it traceable, we need to propagate the exception from the the caller side (e.g. `S3TransferUtil`).


markjschreiber (approved):
Thanks!

PR #647, opened 2025-03-20
Throw `NoSuchFileException` when copying a non-existent file

arouel:
When copying a non-existent S3 object, a `NoSuchFileException` should be thrown. This would make it easier for the caller to handle this case explicitly.


markjschreiber (approved):
Thanks for the update!

PR #649, opened 2025-03-24
Provide a dedicated `IOException` to handle expected failures

arouel:
If we want to re-apply changes on a `FileChannel` or `SeekableByteChannel` that failed to upload due to failing preconditions, we need to figure out if the resulting failure was a `PreconditionFailed`. To figure this out, we would need to parse the stacktrace, the causes and their message. To avoid that, we need a custom `IOException` that enable us to check the status code or error code from the causing `AwsServiceException`.

### DIFF
--- a/src/main/java/software/amazon/nio/spi/s3/S3TransferException.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3TransferException.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.nio.spi.s3;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import software.amazon.awssdk.awscore.exception.AwsServiceException;
+
+public class S3TransferException extends IOException {
+    private static final long serialVersionUID = 1L;
+
+    private final String errorCode;
+    private final String errorMessage;
+    private final Integer numAttempts;
+    private final String requestId;
+    private final int statusCode;
+
+    public S3TransferException(String method, Path path, AwsServiceException cause) {
+        this(
+            errorCode(cause),
+            errorMessage(cause),
+            cause.requestId(),
+            cause.statusCode(),
+            cause.numAttempts(),
+            message(method, path, cause),
+            cause);
+    }
+
+    private S3TransferException(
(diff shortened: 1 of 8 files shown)


src/main/java/software/amazon/nio/spi/s3/S3TransferException.java
```diff
@@ -0,0 +1,81 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.nio.spi.s3;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import software.amazon.awssdk.awscore.exception.AwsServiceException;
+
+public class S3TransferException extends IOException {
```
markjschreiber, line 12: It would be valuable to have javadocs to explain what this exception means and when it should be used. Especially as it is a `public` class

src/main/java/software/amazon/nio/spi/s3/S3TransferException.java
```diff
@@ -0,0 +1,81 @@
...
+        Throwable cause) {
+        super(message, cause);
+        this.errorCode = errorCode;
+        this.errorMessage = errorMessage;
+        this.requestId = requestId;
+        this.statusCode = statusCode;
+        this.numAttempts = numAttempts;
+    }
+
+    private static String errorCode(AwsServiceException cause) {
```
markjschreiber, line 48: These `private static` methods appear to be unused. What are they intended for? Can they be removed?

markjschreiber (changes requested):
- Add a Javadoc comment for the `S3TransferException` class
- There are some private static methods in the Exception that don't appear to be used.

src/main/java/software/amazon/nio/spi/s3/S3TransferException.java
arouel: Agreed, I added some JavaDoc to explain the purpose and how it is meant to be used.

src/main/java/software/amazon/nio/spi/s3/S3TransferException.java
arouel: The method `errorCode` is used on line 38.

markjschreiber (commented):
Looks good. I recently added minimum code coverage requirements for the gradle build so you may need to augment unit tests.

markjschreiber (approved):

PR #654, opened 2025-03-25
Prevent put-object request when nothing has changed

arouel:
We want to avoid unnecessary `PutObject` requests to S3 if the content of the local temporary file has not changed. For this we introduce a new `S3OpenOption` called `S3PutOnlyIfModified` that holds the business logic to calculate a checksum while opening a `FileChannel` or `SeekableByteChannel` and verifying if the content has changed.

To be able to reuse the existing code to calculate checksums for the object integrity check, we restructured `S3OpenOption` to unify the API. This enables us to reuse the existing checksum algorithms within `S3PutOnlyIfModified`.


markjschreiber (approved):
Looks good! Only minor comment is that if you do make a put with the same contents then the created date in S3 will update. If you use this new option then the created date will obviously not update which could surprise someone who didn't know you were using this option.

PR #686, opened 2025-07-31
Support different signers for non AWS S3-Storage

sstamm:
We use S3 provided by NetApp which does not support payload-signing.
To use the AWS S3Client you have to configure the Signer like:
```java
        this.client = S3Client.builder()
            .overrideConfiguration(ClientOverrideConfiguration.builder().putAdvancedOption(SdkAdvancedClientOption.SIGNER, AwsS3V4Signer.create()).build()).build();
``` 

This change introduces ``s3.spi.client.custom-signer`` to define a customer signer class like:

```java

    public static class Signer extends AbstractAwsS3V4Signer    {
        public Signer()        {
            //  Auto-generated constructor stub
        }
    }
...

        System.setProperty("s3.spi.client.custom-signer", Signer.class.getName());
```

### DIFF
--- a/src/main/java/software/amazon/nio/spi/s3/S3ClientProvider.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3ClientProvider.java
@@ -5,13 +5,18 @@
 
 package software.amazon.nio.spi.s3;
 
+import java.lang.reflect.InvocationTargetException;
+import java.time.Duration;
+
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
-import java.time.Duration;
+
 import software.amazon.awssdk.core.client.config.ClientOverrideConfiguration;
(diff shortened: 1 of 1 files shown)


markjschreiber (changes requested):
Thanks for suggesting this change. While reviewing I noticed that the `Signer` class is deprecated and the recommendation is to use the `software.amazon.awssdk.http.auth.spi.signer` package, which is itself an SPI and might allow you to introduce a custom signer without needing to change this package? 
Can you explore this possibility? Unfortunate there doesn't seem to be much documentation on this SPI. https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/http/auth/spi/signer/package-summary.html

sstamm:
I am struggling at this deprication for some time because it's not clear (for me as an S3 newbie) how to adopt it.
I found: https://github.com/aws/aws-sdk-java-v2/issues/5962 
Any idea where I can get more information?

Issue #703, opened 2025-10-23
Token expired when use WebIdentityTokenFileCredentialsProvider after few hours,


ZhengXinCN:
Why not pass the CredentialsProvider instead AwsCredentials in S3NioSpiConfiguration ?

java.io.IOException: software.amazon.awssdk.core.exception.SdkClientException: Failed to send the request: Token expired - needs a refresh. (SDK Attempt Count: 1)
	at software.amazon.nio.spi.s3.S3FileSystemProvider.newDirectoryStream(S3FileSystemProvider.java:373)
	at java.base/java.nio.file.Files.newDirectoryStream(Unknown Source)
	at java.base/java.nio.file.Files.list(Unknown Source)

PR #705, opened 2025-10-30
add AwsCredentialsProvider support

ZhengXinCN:
Issue: (https://github.com/awslabs/aws-java-nio-spi-for-s3/issues/703)

allow pass AwsCredentialsProvider into S3 Client directly, which solve the taken refreshe problem

### DIFF
--- a/src/main/java/software/amazon/nio/spi/s3/S3ClientProvider.java
+++ b/src/main/java/software/amazon/nio/spi/s3/S3ClientProvider.java
@@ -150,9 +150,9 @@ public class S3ClientProvider {
             builder.endpointOverride(endpointUri);
         }
 
-        var credentials = configuration.getCredentials();
-        if (credentials != null) {
-            builder.credentialsProvider(() -> credentials);
+        var credentialsProvider = configuration.getCredentialsProvider();
+        if (credentialsProvider != null) {
+            builder.credentialsProvider(credentialsProvider);
         }
 
         var region = configuration.getRegion();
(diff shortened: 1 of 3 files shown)


src/main/java/software/amazon/nio/spi/s3/S3ClientProvider.java
```diff
@@ -126,9 +126,9 @@ S3CrtAsyncClientBuilder configureCrtClient() {
             asyncClientBuilder.endpointOverride(endpointUri);
         }
 
-        var credentials = configuration.getCredentials();
```
markjschreiber: Deleting this is a potential breaking change. Your later javadoc suggests that both could be provided and credentialsProvider has precedent. If a credentials provider is not supplied then I think people may still want to be able to provide credentials.

src/test/java/software/amazon/nio/spi/s3/config/S3NioSpiConfigurationTest.java
```diff
@@ -214,6 +215,46 @@ public void withAndGetCredentials() {
...
+    public void withAndGetCredentialsProvider() {
+
+        final AwsCredentialsProvider C1 = () -> AwsBasicCredentials.create("key1", "secret1");
+        final AwsCredentialsProvider C2 = () -> AwsBasicCredentials.create("key2", "secret2");
+
+        then(config.withCredentialsProvider(C1)).isSameAs(config);
+        then(config.getCredentialsProvider()).isSameAs(C1);
+        then(config.withCredentialsProvider(C2)).isSameAs(config);
+        then(config.getCredentialsProvider()).isSameAs(C2);
+        then(config.withCredentialsProvider(null).getCredentials()).isNull();
```
markjschreiber: This sets the provider to null and then checks `getCredentials()` is `null`. It is valid as initially no credentials were set, but it's potentially confusing (I would expect to assert `getCredentialsProvider()` is `null` after setting provider to `null`). The test later uses `withCredentialsProvider(null).getCredentialsProvider()` — that second usage is clearer. Consider simplifying/asserting provider-specific getters to reduce confusion.

markjschreiber (approved):
LGTM. Thanks for the contribution!

PR #770, opened 2026-07-23
bug: do not throw exception when bucket specified in URI already exis…

scsosna99:
…ts and is owned by you.

Take the URI s3://myBucket.  The first time newFileSystem is called, the bucket is correctly/successfully created before creating/caching the FileSystem.  Great.

However, all subsequent calls throw an IOException due to the underlying BucketAlreadyOwnedByYouException, thereby requiring the S3 bucket to be deleted between each usage of the file system.  Not ideal, as everything persisted to the file system is lost.

This change allows a bucket to be reused if owned by you, thereby allowing true persistence.

*Issue #, if available:*

*Description of changes:*


markjschreiber:
The nio contract for the `[newFileSystem](https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileSystems.html#newFileSystem-java.net.URI-java.util.Map-)` method suggests that the expected behavior is to create a filesystem and return a `FileSystemAlreadyExistsException` if the file system has already been created.

To get a reference to an existing s3 bucket you should call `getFilesystem(URI)`.

Do you have control of the application that is calling `newFilesystem`? If so I would recommend catching the exception there and returning the result of `getFilesystem(URI)` as needed.

If you don't you might want to file an issue with the owner of the package as I think they are misusing the `newFilesystem` method.

scsosna99:
My interpretation, based on how zip/jar file systems works is that the `FileSystemAlreadyExistsException` is thrown when the file system has already been loaded into the JVM, not that it exists or not.  In the zip/jar example, this prevents people from loading and reloading the same file again again; rather, you load it once (`newFileSystem`) and subsequently retrieve it (`getFileSystem`) and then, within the running JVM only one instance of the `FileSystem` object is loaded.

Therefore, I assumed `S3FileSystemProvider` would only create new when the file system has not already been loaded into the JVM; the attempt to create the bucket is cool, I'm only suggesting that if the bucket already exists AND it is owned by you, then it moves on and creates the file system.

The `S3FileSystemProvider.getFileSystem` is also incomplete because it assumes that the underlying bucket is present; however, if it isn't, it just magically goes on.  If we follow your interpretation, then get should either create a bucket as well or throw an exception if the bucket doesn't exist.

Reading the JavaDocs for `[FileSystemProvider.newFileSystem`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/file/spi/FileSystemProvider.html#newFileSystem(java.net.URI,java.util.Map)) is _The method throws FileSystemAlreadyExistsException if he file system already exists because it was previously created by an invocation of this method.  Once a file system is closed it is provider-dependent if the provider allows a new file system to be created with the same URI as a file system it previously created._ which seems to say both our interpretations are correct.

I (obviously) prefer my interpretation because it means a solution doesn't have to be cognizant of whether the file system was _ever_ created by _anyone_, just need to track what you have done.  But apparently anything is legit....

scsosna99:
Looking for other examples, Azure Blob does the same as the jar/zip provider, keeping track of file systems instantiated within this running JVM and tracking them in a ConcurrentHashMap.

```
  public FileSystem newFileSystem(URI uri, Map<String, ?> config) throws IOException {
        String endpoint = extractAccountEndpoint(uri);

        if (this.openFileSystems.containsKey(endpoint)) {
            throw LoggingUtility.logError(ClientLoggerHolder.LOGGER,
                new FileSystemAlreadyExistsException("Name: " + endpoint));
        }

        AzureFileSystem afs = new AzureFileSystem(this, endpoint, config);
        this.openFileSystems.put(endpoint, afs);

        return afs;
    }
```

Google Cloud, if I'm reading the code correctly (I have no experience with GC) just blindly creates the file system regardless of whether it's a dupe or not.

markjschreiber:
I think you might be right. I took a look the the `jdk.nio.zipfs.ZipFileSystemProvider` from OpenJDK and it looks like this:

```
  public FileSystem newFileSystem(URI uri, Map env) throws IOException {
      Path path = uriToPath(uri);
      synchronized (filesystems) {
          Path realPath = null;
          if (ensureFile(path)) {
              realPath = path.toRealPath();
              if (filesystems.containsKey(realPath))   // <-- in-JVM tracking map
                  throw new FileSystemAlreadyExistsException();
          }
          ZipFileSystem zipfs = getZipFileSystem(path, env);
          if (realPath == null) realPath = path.toRealPath();
          filesystems.put(realPath, zipfs);
          return zipfs;
      }
  }
```

And for `sun.nio.fs.UnixFileSystemProvider` the implementation is:

```
  public final FileSystem newFileSystem(URI uri, Map env) {
      checkUri(uri);
      throw new FileSystemAlreadyExistsException();   // the single default FS already exists at startup
  }
  public final FileSystem getFileSystem(URI uri) {
      checkUri(uri);
      return theFileSystem;
  }
  ```


To align with these references, the "already exists" decision should be based on `FS_CACHE`, not on the S3 bucket state:

  - `newFileSystem`: if `FS_CACHE` already contains an entry for the key, `throw FileSystemAlreadyExistsException; ` otherwise create the `S3FileSystem`, register it, and return it. Treating `BucketAlreadyOwnedByYouException` as success (proceed to register the FS) is the correct fix — it mirrors ZipFS opening an already-existing backing file. `BucketAlreadyExistsException` (owned by someone else)
  should remain an error, but arguably an `IOException`/access error rather than `FileSystemAlreadyExistsException`.
  - The computeIfAbsent-swallowing-duplicates behavior should be replaced with an explicit contains-check so the "already exists" contract is actually enforceable.

scsosna99:
This PR likely should be expanded: the `getFileSystem` should be changed to _not create the file system if it doesn't exist and the `newFileSystem` then needs to check the cache and throw an exception if it hasn't already been new'ed.

I am willing to make the wider changes, unfortunately this feels non-backwards-compatible.  The implementation changes themselves are minor but the change to `getFileSystem` breaks 100s of tests (I started down this path and decided to minimize my work for the time being).

Let me know how you'd prefer to proceed.  Thanks.

markjschreiber:
Yes, strictly speaking `getFileSystem` would also need to change and as you say it would be a breaking change. I have been contemplating a move to a new major version to redesign some other issues that would also be breaking and do a full audit of the implementation against the contract/ references. Idealy a new and corrected `getFileSystem` implementation should go there. However, without making the breaking change the interaction of this incorrect `getFileSystem` with a corrected `newFileSystem` would introduce new potential problems. It introduces an active new failure mode where ordinary path usage makes `newFileSystem` throw and skip bucket provisioning.

Given the above I think a breaking change is required (along with the necessary test and example code updates).

Regarding the proposed fix in this PR, `newFileSystem` should also consult the `FS_CACHE` to mimic the "is this filesystem tracked by the JVM". In the current code: the duplicate call at throws `FileSystemAlreadyExistsException` (for the wrong reason); in your proposed change, the duplicate call returns the cached instance and never throws. This is also different from `ZipFileSystemProvider`, where a second `newFileSystem` for an already-registered path throws immediately via `filesystems.containsKey(realPath)`. So that would need to be modified.
  
The solution would also need to cautiously handle `InterruptedException | TimeoutException | SdkException e` making sure everything is properly cleaned up by closing any new FS that was created/ potentially created and removing it from the `FS_CACHE`.

Let me know if you want to take a crack at that, or if you'd prefer I do it?

scsosna99:
I'd love to take a shot at it next week, I won't have time this weekend but by Tuesday I should have a chance to dig in.  Thanks

markjschreiber:
> I'd love to take a shot at it next week, I won't have time this weekend but by Tuesday I should have a chance to dig in.


Excellent, looking forward to it!

scsosna99:
Pushed the updated PR, I believe I addressed your concerns but let me know if there's other changes to include.  Thanks.

scsosna99:
@markjschreiber I see you've been busy on the v3 that breaks compatibility, which likely explains why this PR hasn't moved along.  Would you like me to focus on your draft PR, review your goals, etc., or something else?  Willing to help if that moves things along (and, selfishly, helps me with my work).  Thanks!

markjschreiber:
> @markjschreiber I see you've been busy on the v3 that breaks compatibility, which likely explains why this PR hasn't moved along. Would you like me to focus on your draft PR, review your goals, etc., or something else? Willing to help if that moves things along (and, selfishly, helps me with my work). Thanks!

Yes, the goal was to make all the breaking changes at once to bring the project into line with the NIO contract. Please let me know if you see any gaps or problems.

